PR #61, opened 2016-05-11
Pchr 1006 pivot table

brunoqueiros:
In this PR was made some CSS changes to update the pivot table style and was created a new component (custom-checkbox), a mixin (placeholder-color) and added the scrollbar fallback.

Before
image

Final result
image
## Placeholder color

It was create a mixin to change the placeholder text color.
`@include placeholder-color(red);`
## Custom checkbox

A new component was created to style the checkbox. The class needs to be in a wrapper element because [Firefox doesn't style element without content](http://stackoverflow.com/a/12833498) `e.g: checkbox`. So to style an element is necessary to have a similar markup.

```
<div class="chr_custom-checkbox">
    <input type="checkbox" id="test" />
    <label for="test">Test</label>
</div>
```
## Filter box

As the pivot table library doesn't support custom templates, some elements were hidden via CSS and another were made via JS in this [PR](https://github.com/compucorp/civihr-employee-portal/pull/122).

**Original**
image

**After changes**
image


danilo-compucorp:
Looks ok. PR approved.

PR #62, opened 2016-05-19
Pchr 1005 filters section

brunoqueiros:
**Before**
before

**After**
after

It was made some changes to let the panel according to PSD and created a class `panel--sliding-body` to sliding the filter panel.


civihr_default_theme/assets/sass/partials/_page.scss
```diff
@@ -15,4 +15,45 @@
         padding: 0;
         background-color: transparent;
     }
+
+    .pane-content {
+      overflow: visible !important;
+    }
+
+    #bootstrap-theme .dropdown-menu[datepicker-popup-wrap] {
+      width: 316px;
```
AkA84, line 24: This is weirdly specific, how so?

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -3,14 +3,27 @@
...
+
+      .panel-body-wrap {
+        max-height: 0;
+      }
+    }
+}
+
+.panel-body-wrap {
+  transition: max-height .4s linear;
+  max-height: 162px;
```
AkA84, line 14: Very specific value, what's the reason?

civihr_default_theme/assets/sass/partials/_page.scss
brunoqueiros, line 24: Because without this value the width will be the parent's size. This value `316px` it's the value at _hrjobcontract_

civihr_default_theme/assets/sass/partials/_panel.scss
brunoqueiros, line 14: It's just a value higher than the `.panel-body-wrap` original height, it's used to work with CSS transition.

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -3,14 +3,27 @@
 
 .panel {
     @include box-shadow($box-shadow-global);
+
+    &.panel--sliding-body {
```
danilo-compucorp: Why not `&--sliding-body` ?

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -3,14 +3,27 @@
 
 .panel {
     @include box-shadow($box-shadow-global);
+
+    &--sliding-body {
+      overflow: hidden !important;
+
+      .panel-body-wrap {
```
danilo-compucorp: If we're using BEM (at least on the new styles), we shouldn't nest selectors

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -18,6 +14,10 @@
   max-height: 162px;
 }
 
+.panel-body-wrap--collapse {
```
danilo-compucorp: Why not:
`.panel-body-wrap {
  transition: max-height .4s linear;
  max-height: 162px;

  &--collapse {
    max-height: 0;
  }
}` 
?

danilo-compucorp:
PR approved

PR #63, opened 2016-05-24
PCHR-1004: Reports list page

brunoqueiros:
It was styled the reports list page to be like the PSD and added a collapse effect on 'Reports Settings' block.

**Before**
image

**After**
anim
## _page.scss

It was created a selector `.ta-hidden-input` to hidden the text angular hidden input, the `bootstrap-civicrm-style.css` already have this selector but it's inside the namespace `#bootstrap-theme`, so to avoid use JS to add a wrap selector I created the selector globally.
## _panel.scss

It was created some selectors to adjust the page style with the PSD.
### .chr_panel--without-background

Removes the panel background and let transparent and set the border color darker than the background.

image

``` html
<div class="panel panel-pane pane-block chr_panel--no-padding chr_panel--without-background">
    <div class="pane-content">
        <div class="chr_search-result__header chr_panel__header">
            <div class="chr_search-result__total">
                <i class="chr_search-result__icon glyphicon glyphicon-chevron-down"></i>
                Panel title
            </div>
        </div>

        <div class="panel-body">
           Content
         </div>
    </div>
</div>
```
### .chr_panel__header__subtitle

Style the subtitle with a different color (link color)

image

``` html
<div class="panel panel-pane pane-block chr_panel--no-padding">
    <div class="pane-content">
        <div class="chr_search-result__header chr_panel__header">
            <div class="chr_search-result__total">
                <i class="chr_search-result__icon glyphicon glyphicon-chevron-down"></i>
                Panel title: <span class="chr_panel__header__subtitle">subtitle</span>
            </div>
        </div>

        <div class="panel-body">
           Content
         </div>
    </div>
</div>
```
## block--menu--hr-reports-settings.tpl.php

It was overridden the markup, and settled this block as an angular controller.


AkA84:
@brunoqueiros the transition doesn't really look good, i see there is some shifting on the left, and the logo gets pulled down for no apparent reason, isn't there a way to improve it?

brunoqueiros:
The reason is because the scrollbar is activated, to improve this I can set to the scrollbar always be visible.

AkA84:
Oh right, then there's no problem

AkA84:
@brunoqueiros i see that you're adding a couple of components variations like `.chr_panel--without-background` and `.chr_panel__header__subtitle`. Whenever you do that, you have to provide in the PR description examples of how to use them (with code snippets) and what the end result is (with a screenshot), so that these additions can be justified. Please see [here](https://github.com/compucorp/civihr-employee-portal-theme/pull/14) for reference:

civihr_default_theme/assets/sass/partials/_page.scss
```diff
@@ -61,4 +61,8 @@
       margin-right: 20px;
       padding-right: 0 !important;
     }
+
+    .ta-hidden-input {
```
AkA84, line 65: What's the `ta` prefix?

brunoqueiros, line 65: It's from [this module](http://textangular.com/)

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -88,6 +88,15 @@
     position: relative;
 }
 
+.chr_panel--without-background {
```
danilo-compucorp: Why not use the "SASS ampersand" in the parent selector block (.chr_panel) ?

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -88,6 +88,15 @@
...
+   background: none;
+   box-shadow: none;
+
+   .chr_search-result__header--border-darker {
```
danilo-compucorp: Should this be following the BEM convention?
If so, then the selectors shouldn't be nested. Please refer to [the BEM faq](http://getbem.com/faq/#css-nested-elements)

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -196,6 +205,10 @@ form {
     }
 }
 
+.chr_panel__header__subtitle {
```
danilo-compucorp: Please refer to [the BEM faq](http://getbem.com/faq/#css-nested-elements)

civihr_default_theme/assets/sass/partials/_panel.scss
brunoqueiros: As I talked with @AkA84, as `chr_search-result__header--border-darker` is related with `.chr_panel--without-background`, it's better to be nested.

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -196,6 +205,10 @@ form {
     }
 }
 
+.chr_panel__header-subtitle {
```
danilo-compucorp: The SASS ampersand could be used here as well, inserting this into the parent block ".chr_panel"

civihr_default_theme/assets/sass/partials/_panel.scss
```diff
@@ -129,6 +129,10 @@ form {
 
 .chr_panel__header {
     box-shadow: none;
+
+    &__header-subtitle {
```
danilo-compucorp: For this to get compiled as `.chr_panel__header-subtitle` it should be written as `.&__header-subtitle` inside its [parent block](https://github.com/compucorp/civihr-employee-portal-theme/blob/33967de47b2eb7691163e1ce1c5d4fd2df03bb6c/civihr_default_theme/assets/sass/partials/_panel.scss#L87)

danilo-compucorp:
Approved

PR #66, opened 2016-06-14
PCHR-1167 - Modifying apply_bootstrap() function to include #prefix and #suffix webform attributes.

coldrunKacper:
- Modifying apply_bootstrap() function to handle #prefix and #suffix field attributes.
- Moving apply_bootstrap() function into template.php so it can be reachable from SSP module.


omarabuhussein:
Approved

PR #71, opened 2016-07-11
PCHR-1005: Fix datepicker and header divider line

brunoqueiros:
This PR removes the gray line which divider the filters header when it's closed, it was updated the datepicker selector to works with the newest version, also was added a border on the calendar icon.

**_Before**_
image

image

**_After**_
image

image


civihr_default_theme/assets/sass/partials/_page.scss
```diff
@@ -20,8 +20,12 @@
       overflow: visible !important;
     }
 
-    #bootstrap-theme .dropdown-menu[datepicker-popup-wrap] {
-      width: 316px;
+    #bootstrap-theme {
+      display: inline-block;
```
AkA84, line 24: `#bootstrap-theme` it's just the namespace, i don't think we should really apply any style to it. What's the purpose of this `inline-block`? Can't you apply it to some other element?

civihr_default_theme/assets/sass/partials/_page.scss
```diff
@@ -20,8 +20,12 @@
...
+
+      [uib-datepicker-popup-wrap] .dropdown-menu {
+        width: 316px;
```
AkA84, line 27: This seems extremely specific, what's it for?

brunoqueiros, line 27: To set the calendar width. https://github.com/compucorp/civihr-employee-portal-theme/pull/62#discussion_r64066874

civihr_default_theme/assets/sass/partials/_page.scss
brunoqueiros, line 24: Because the `#bootstrap-theme` element is added between the input and the icon, so without the `display: inline-block;` the layout break.

image

civihr_default_theme/assets/sass/partials/_page.scss
brunoqueiros, line 27: These are the changes that makes necessary to specify the dropdown width.

Without this one apparently the only thing that breaks is the pivot table so we can move this rule to `partials/_pivot-table.scss`.
https://github.com/compucorp/civihr-employee-portal-theme/blob/develop/civihr_default_theme/assets/sass/civihr_default/_custom.scss#L199
Commit: https://github.com/compucorp/civihr-employee-portal-theme/commit/e187a5a94ccf3f019eb0a35a0384c41a1bf3d04f

I didn't find anything which will breaks without this rule, needs to check.
https://github.com/compucorp/civihr-employee-portal-theme/blob/develop/civihr_default_theme/assets/sass/civihr_default/_custom.scss#L502
Commit: https://github.com/compucorp/civihr-employee-portal-theme/commit/180bed47d9ed72e2f3dd3668aed49af206cd8198

AkA84, line 27: Ok, for now it's fine but definitely all this stuff will be refactored

AkA84:
Approved

PR #73, opened 2016-07-18
PCHR-1292 - Implementing dynamic variable 'container_class' into page tpl.php (handling a different class on Reports pages).

coldrunKacper:
Creating additional variable for page.tpl.php which can set different div class on Reports page.


coldrunKacper:
Closed as it was wrongly pointing to staging instead of develop.

PR #74, opened 2016-07-18
Pchr 1292 civicrm layout broken

coldrunKacper:
Creating additional variable for page.tpl.php which can set different div class on Reports page.


civihr_default_theme/template.php
```diff
@@ -38,6 +38,23 @@ function civihr_default_theme_preprocess_page(&$variables) {
...
+  $variables['container_class'] = $container_class;
+}
+
+/**
+ * Check if we are displaying custom Report page.
+ *
+ * @return boolean
+ */
+function _is_hrreports_current_path() {
+    if (substr(current_path(), 0, 8) === 'reports/') {
```
brunoqueiros: Could this be `return (substr(current_path(), 0, 8) === 'reports/');`

AkA84: Watch out the indentation spaces also

brunoqueiros:
Approved!

PR #79, opened 2016-07-22
PCHR-1007: Overrides the pager's markup

brunoqueiros:
This PR overrides the pager's markup to looks like the design, it was hidden the `first` and last `buttons` and shows the first page and the last page after the `...`

**_Before**_
anim

**_After**_
pchr-1007-pager


coldrunKacper:
Approved.

PR #80, opened 2016-07-25
PCHR-1337: Rename the reports settings file

brunoqueiros:
Recently the "Reports settings" block was slightly changed (the Drupal machine name was changed) so there is a need to refactor one of current Theme template (change its name).

The Theme repo file name should be changed:
**_from:**_ `templates/block/block-menu-hr-reports-settings.tpl.php`
**_to:**_ `templates/block/block-menu-menu-reports-settings.tpl.php`


coldrunKacper:
Approved.

PR #81, opened 2016-08-05
PCHR-1385: Remove ARRAY_FILTER_USE_BOTH usage for php 5.5 compatibility

omarabuhussein:
## Problem

the **array_filter** core php method was used with **ARRAY_FILTER_USE_BOTH** option which is introduced in PHP 5.6.0 , which  cause the minimum PHP  version to run the Civihr/SPP to be 5.6.0 too . 

But as specified here https://wiki.civicrm.org/confluence/display/CRMDOC/CiviCRM+PHP+Requirements , the minimum version should be 5.5 .
## Solution

I replaced the array_filter with ARRAY_FILTER_USE_BOTH by **foreach** implementation.

### DIFF
--- a/civihr_default_theme/template.php
+++ b/civihr_default_theme/template.php
@@ -296,9 +296,13 @@ function civihr_default_theme_form_element_label($variables) {
  * @return array
  */
 function bootstrapable_fields($fields_structure) {
-  return array_filter($fields_structure, function($value, $key) {
-    return substr($key, 0, 1) <> '#' && !in_array($value['#type'], ['hidden']);
-  }, ARRAY_FILTER_USE_BOTH);
+  $fields = $fields_structure;
+  foreach($fields as $key => $value)  {
+    if (!(substr($key, 0, 1) <> '#' && !in_array($value['#type'], ['hidden'])))  {
+      unset($fields[$key]);
+    }
+  }
+  return $fields;
 }
 
 /**


civihr_default_theme/template.php
```diff
@@ -167,9 +167,12 @@ function civihr_default_theme_form_element_label($variables) {
  * @return array
  */
 function bootstrapable_fields($fields_structure) {
-  return array_filter($fields_structure, function($value, $key) {
-    return substr($key, 0, 1) <> '#' && !in_array($value['#type'], ['hidden']);
-  }, ARRAY_FILTER_USE_BOTH);
+  foreach($fields_structure as $key => $value)  {
+    if (!(substr($key, 0, 1) <> '#' && !in_array($value['#type'], ['hidden'])))  {
+      unset($fields_structure[$key]);
```
AkA84: Question: i believe that [array_filter](http://php.net/manual/en/function.array-filter.php) returns a _new_ array (you might want to confirm it), but here you're modifying the existing one. Could you make sure that the original `$fields_structure` remains unchanged?

omarabuhussein: the value is not passed by reference to this function so it doesn't matter in this case ... but I will do it just case there is some hidden drupal magic that I am not aware of .

AkA84:
Approved

PR #86, opened 2016-08-09
PCHR-1403: Delete civihr_admin_theme

AkA84:
The civihr_admin_theme was created for the old reports, as a way to mimic the look and feel of the civihr theme on the SSP.

Given that we now have a new report system which utilizes the ssp theme, there is no need for this anymore, and it can then be deleted


omarabuhussein:
approved

coldrunKacper:
Approved.

PR #93, opened 2016-08-18
PCHR-1249: Outer page HTML+CSS

deb1990:
## As part of this ticket the following things are added in My Details page of SSP
1. Title
2. Actions button
3. Tabs

image
## For this changes to work
1. Create a new menu item for My Leave
2. Point the same to **/myleaves**
3. Create a page and put path as "**/myleaves**", set title as "**My Leaves**"
4. Create a view called "**myleave**"
5. Add a block in the same, also delete the default page for the view.
6. Put "**myleave_block**" as the machine name for the block. 
7. Goto "**admin/structure/block/manage/views/myleave-myleave_block/configure**"
8. Add "**myleaves**" under "**Only the listed pages**"


civihr_default_theme/assets/partials/myleave.html
```diff
@@ -0,0 +1,50 @@
+<div>
+
+  <div class="button-container">
+
```
AkA84: Please remove all of these whitespaces

civihr_default_theme/assets/partials/myleave.html
```diff
@@ -0,0 +1,50 @@
...
+    <div class="btn-group">
+      <button type="button" class="chr_action dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
AkA84: Don't use `.chr_action` or **any** of these custom components that could be traced back to a Bootstrap component instead. So instead of `.chr_action`, this is a `.btn.btn-primary`

civihr_default_theme/scss/components/_my-leaves.scss
```diff
@@ -0,0 +1,3 @@
+.button-container{
```
AkA84: This should be in _scss/layout/_page.scss_ or _scss/layout/_my-leaves.scss_ or _scss/pages/_my-leaves.scss_ (whatever you prefer)

Also remember to namespace this selector with the class associated with this specific page

civihr_default_theme/scss/components/_tabs.scss
```diff
@@ -0,0 +1,38 @@
...
+        color: $brand-primary;
+        border: none;
+        border-bottom: 3px solid $brand-primary;
+      }
+    }
+  }
+}
+
+
+.nav, .chr_table-w-filters__filters__nav{
```
AkA84: Every selector go in a new line (and put a whitespace before the `{`)

``` scss
.nav,
.chr_table-w-filters__filters__nav {
```

civihr_default_theme/templates/page/page--myleaves.tpl.php
```diff
@@ -0,0 +1,113 @@
+<?php
```
AkA84: This is the template of the _entire_ page, but you should only provide the template of the page content

PR #95, opened 2016-08-19
PCHR-1249: Outer page HTML+CSS

deb1990:
Original PR: #93 
## As part of this ticket the following things are added in My Details page of SSP
1. Title
2. Actions button
3. Tabs

image
## For this changes to work
1. Create a new menu item for My Leave
2. Point the same to **myleave**
3. Create a page and put path as "**/myleave**", set title as "**My Leave**"
4. Create a view called "**myleave**"
5. Add a block in the same, also delete the default page for the view.
6. Put "**myleave_block**" as the machine name for the block. 
7. Goto "**admin/structure/block/manage/views/myleave-myleave_block/configure**"
8. Add "**myleave**" under "**Only the listed pages**"
## Other Changes:
1. CSS from **sass/_utils.scss** moved to **scss/base/_typography.scss** and _utils.scss has been **deleted**.
2. CSS for Tabs optimised.
3. **btn-default** CSS modified to keep same styling across all default buttons. All Tab related CSS now is now done only using ".nav" selectors instead of using **btn-default**.

#### Problem
By default a pane (which now contains the my leave page content) has a white background, box shadow around and a padding
before-pane
#### Solution
The specific pane of this page has now custom css that counteract the default styling
after-pane
#### Problem
We were trying to apply layout rules by selecting the outer cms markup.
```scss
.block-civihr-leave-absences {

  #my-leave-block {

  }
}
```
The problem with that is that if we're using a different cms, or if the page gets put inside a different container, the styling will not be applied anymore
#### Solution
The page got a custom `.my-leave-page` class that is now used to apply layout rules
```scss
.my-leave-page {

  .button-container {
    margin-bottom: 40px;
  }
}
```
#### Problem
There were several problems with this partials

1. We were keeping our custom values for the variables on the top of the file, while keeping the original value below. This was done with the intention to keep a reference to the original value, but it is unnecessary given that the main scss file (*civihr_default_theme.style.scss*) includes *both* this file and the original *_variable.scss* partials (see below). For this reason this partial should contain only the custom values
```scss
// Bootstrap global partials.
// -----------------------------------------------------------------------------
@import "base/variables";

// Core variables and mixins
@import "bootstrap/variables";
```

2. Some of the custom variables (any `$crm_` variable) were used only once or were unnecessary. For example `$crm-brand-tertiary` was equal to `$brand-primary`, making it kinda pointless

3. The `$crm_` prefix for the custom variables does not match the `$crm-` prefix used for the custom variables in the CiviHR theme

#### Solution
1. The variables order now respects the one in the original partial, and only the custom variables or the original variables with custom values are now listed in it

2. Some of the variables have been removed, other have been renamed with a more meaningful name

3. The prefix used now is `$crm-`


civihr_default_theme/assets/partials/myleave.html
```diff
@@ -0,0 +1,39 @@
+<div id="myleave_block">
```
AkA84: Do you actually need this? Given that you are using this only for applying the padding, can't you use one of the elements that Drupal already provides you? Like `.content` or `.main-container` or the like

If you indeed need this, then please follow the same dasherized naming convention: `.my-leave-block`

deb1990: As discussed keeping this ID. But changed the name to .my-leave-block.

civihr_default_theme/scss/base/_custom.scss
```diff
@@ -278,22 +278,6 @@ body {
     }
 }
 
-/**
- * Customization of the standard Bootstrap button group
- */
-.btn-group {
```
deb1990, line 284: Removed this code as we are changing the look of Navs using ".nav" classes which makes this code unnecessary.

civihr_default_theme/scss/base/_variables.scss
```diff
@@ -34,6 +34,7 @@ $gray                 : #767b8e; // main text/lighter, menu item bg hover, foote
 $gray-dark            : #555862; // menu bg, default: lighten(#000, 20%) !default
 $gray-darker          : #4c4c5d; // text darker, default: lighten(#000, 13.5%) !default; // #222
 $crm_gray-darker-alt  : #404454; // page title
+$crm_gray-border      : #777C8F; // page title
```
AkA84: The name of the variable isn't good, it's referring to a specific element (`border`) while by following bootstrap convention it should convey only the color (`gray`) and if necessary a specific shade of it (`darker`, `light`, etc)

This name is kinda pointless also by the fact that you're later using it for things that are not a border:

``` scss
.btn-default {
  border-color: $crm_gray-border;
  color: $crm_gray-border;

  &.active {
    border-color: $crm_gray-border;
    background: $crm_gray-border !important;
    color: $light !important;
  }
}
```

so no sense in calling it that. Please give it an appropriate name

civihr_default_theme/scss/pages/_my-leave.scss
```diff
@@ -0,0 +1,10 @@
+.page-myleave {
```
AkA84: Since you decided to keep `#my-leave-block`, I think there's no use using `.page-myleave` as the namespace of this scss partial, you can use `#my-leave-block` directly:

``` scss
#my-leave-block {
  padding: 0 15px;

  .button-container {
    margin-bottom: 40px;
  }
}
```

civihr_default_theme/scss/pages/_my-leave.scss
```diff
@@ -0,0 +1,10 @@
...
+
+  #my-leave-block {
+    padding: 0px 15px;
```
AkA84: If the value is `0` then please don't use any unit

civihr_default_theme/scss/partials/_button.scss
```diff
@@ -0,0 +1,15 @@
+.btn-default {
+  border-color: $crm_gray-border;
+  color: $crm_gray-border;
+
+  &.active {
+    border-color: $crm_gray-border;
+    background: $crm_gray-border !important;
```
AkA84: Why the `!important`? Is it really necessary? What are you overriding?

civihr_default_theme/scss/partials/_navs.scss
```diff
@@ -1,6 +1,10 @@
 // Styles for Navs
 // --------------------------------------------------
 
+.nav > li > a {
+  background-color: $light;
```
AkA84, line 5: This partial is pretty small, can you fix the indentation for the whole of it?

civihr_default_theme/scss/base/_variables.scss
deb1990: Removed $crm_gray-border completely and using **$grey** instead, as the color difference is not significant.

civihr_default_theme/scss/partials/_button.scss
deb1990: It was added to override the hover state color when item is active. Modified to
`&.active,
  &.active:hover {`

civihr_default_theme/scss/pages/_my-leave.scss
```diff
@@ -0,0 +1,10 @@
+#my-leave-block {
+  padding: 0 15px;
+
+  .button-container {
+    margin-bottom: 40px;
+  }
+}
+
+
+
```
AkA84: Empty lines

AkA84:
Approved

civihr_default_theme/scss/base/_custom.scss
```diff
@@ -386,7 +370,12 @@ a {
 
 /* Headings */
 
-h1 {
+.block__title {
```
AkA84, line 373: Where is this class used? I don't see it anywhere

civihr_default_theme/scss/pages/_my-leave.scss
```diff
@@ -0,0 +1,10 @@
+.block-civihr-leave-absences {
```
AkA84: I don't see this being used anywhere, probably because now the template resides somewhere else than your original block?

civihr_default_theme/scss/partials/_button.scss
```diff
@@ -0,0 +1,16 @@
...
+  &.active,
+  &.active:hover {
+    border-color: $gray;
+    background: $gray;
+    color: $light;
+  }
+}
+
+.btn-primary{
+  border-color: transparent;
```
AkA84: I think this should be a variable value like `$btn-primary-color` (although it's likely that we want the border color to be transparent for every contextual variation of a button - success, info, etc -)

civihr_default_theme/scss/partials/_button.scss
```diff
@@ -0,0 +1,16 @@
+.btn-default {
+  border-color: $gray;
+  color: $gray;
```
AkA84: `border-color` and `color` are covered by the following sass variables: `$btn-default-color`, `$btn-default-bg`

civihr_default_theme/scss/partials/_button.scss
```diff
@@ -0,0 +1,16 @@
...
+
+  &.active,
+  &.active:hover {
+    border-color: $gray;
+    background: $gray;
+    color: $light;
```
AkA84: Is it really necessary to redefine all these properties? I tried on my local to have a `.btn.btn-default.active` without those properties and it doesn't seem to look broken

AkA84 (changes requested):

AkA84:
screen

The page doesn't look right. It has the white background and two titles. I've looked into it and it's because of the `.pane-title` and `.pane-content` blocks that get added by drupal. Please check if it's something that we can get rid of via config (which we can then export) or if it's something that we have to override via CSS (first option would be preferrable)

civihr_default_theme/scss/base/_custom.scss
deb1990: This class is for the title of the block which is generated by Drupal.

civihr_default_theme/scss/pages/_my-leave.scss
deb1990: This is the class for the block, autogenerated by Drupal, so kept the namespace.

civihr_default_theme/scss/base/_custom.scss
AkA84: I've never seen Drupal using the BEM convention, nor I could find this class in the actual page

civihr_default_theme/scss/pages/_my-leave.scss
AkA84: But I can't see the block anymore, and in fact (see the screenshot below) the margin-bottom on the button is not applied)

civihr_default_theme/scss/partials/_navs.scss
```diff
@@ -1,28 +1,34 @@
...
-                color: $gray-darker;
-                border: 0;
-                border-bottom: 3px solid transparent;
-            }
+    li {
+      a {
+        font-size: 18px;
+        color: $gray-darker;
+        border: 0;
+        border-bottom: 3px solid transparent;
```
deb1990, line 18: Add `rgba(0, 0, 0, 0.035)` and `transparent` in a variable.

AkA84: @deb1990 sorry not sure i understand what you mean, could write me an example snippet?

deb1990: @AkA84 I mean use a variable, like,
`$crm-box-shadow-color: rgba(0, 0, 0, 0.035)` , 
`box-shadow: 0 3px 10px 0 $crm-box-shadow-colo, 0 4px 8px 0 $crm-box-shadow-color`

Similar for `border-bottom` color

deb1990:
Approved

PR #97, opened 2016-09-03
PCHR-1514: Menu switcher updates

deb1990:
# Requirements:

**SSP:**
1. Combine Logo and Home link into one link with logo AND 'Home' word also remove Home menu.
2. On hover on new link show a dropdown with links:
CiviHR admin - link to /civicrm
CiviHR SSP - link to dashboard
3. For when a user has 'administer content' and 'administer users' permissions show a gear icon on the right.
4. On hover on the gear icon show dropdown with links:
Manage documents - /admin/content 
Manage users - /admin/people
# Please Note

**CiviCRM Administration theme** needs to be changed to **CiviHR Default Theme** to replace the CiviCRM logo
# Screencasts:

**Desktop**
desktop

**Mobile**
mobile


civihr_default_theme/templates/page/page.tpl.php
```diff
@@ -56,6 +68,17 @@
       </ul>
     </div>
     <?php } ?>
+    <?php if (user_access("administer users") && user_access("administer nodes") ) { ?>
+    <div class="chr_header__settings-menu">
+      <span class="icon-settings">
+        <i class="fa fa-cog" aria-hidden="true"></i>
+      </span>
+      <ul class="chr_header__settings-menu__sub-menu">
```
danilo-compucorp: Is this supposed to follow the BEM convention?

deb1990: In this file previously BEM convention was followed, like "chr_header__brand chr_brand". So kept the same.

danilo-compucorp: I prefer to properly use BEM. Please refer to this: http://getbem.com/faq/#css-nested-elements

civihr_default_theme/scss/partials/_menu.scss
```diff
@@ -62,3 +62,98 @@
...
+      > i {
+        margin-right: 10px;
+      }
+    }
+  }
+}
+
+.chr_header__home-menu {
+
+  .chr_header__home-sub-menu {
```
danilo-compucorp: Why nesting this?

deb1990: sub-menu is always supposed to be inside menu, so same structure should be maintained in  SCSS also.

danilo-compucorp: Not necessarily, especially when using BEM

deb1990: Sorry if I dont understand this properly. BE talks about having unique name for elements to 
1. Avoid long names
2. Having the ability the move elements without worrying about changing the name of classes.

But in this case,
1. We avoided long names by having unique name.
2. But as sub menu is never going to be outside menu, I think we should have it nested inside the menu.

Let me know your views on this.

danilo-compucorp: I'm not talking about the names. I'm talking about nesting selectors, and I believe it should be done only when necessary, not for the sake of mimicking the DOM

danilo-compucorp:
the ".orig" files shouldn't be commited

danilo-compucorp:
Approved

ghost:
I think the desktop version works fine. I'm not sure that the mobile version is super clear as there's no real call to action indicating the user should click the logo. Perhaps these links would be better placed inside the menu itself?

mi-kon:
I agree, we don't need a separate dropdown for logo. On click on logo let's open the menu under the burger icon with those two links moved to the first dropdown

deb1990:
Removal of Home menu is taken care from CSS

civihr_default_theme/templates/page/page.tpl.php
```diff
@@ -12,6 +12,11 @@
  * NOTE: The #outer-wrapper element is for the mobile menu to have a way
  * both to stick at the top of the page and to be as tall as the content (necessary because of the Drupal toolbar)
  */
+
+$admin_link = l(t('CiviHR admin'), 'civicrm', array('html' => true));
+$ssp_link = l(t('CiviHR SSP'), 'dashboard', array('html' => true));
+$documents_link = l(t('Manage documents'), 'admin/content', array('html' => true));
+$users_link = l(t('Manage users'), 'admin/people', array('html' => true));
```
AkA84: missing indentation on all 4 lines

civihr_default_theme/templates/page/page.tpl.php
```diff
@@ -20,15 +25,22 @@
       <div class="chr_header__nav__toggle">
         <i class="fa fa-2x fa-navicon"></i>
       </div>
-      <div class="chr_header__corner__brand chr_brand">
-        <a class="chr_brand__link" href="<?php print rtrim($base_path, '/') ?>/dashboard" title="<?php print htmlspecialchars($site_name); ?>">
+      <div class="chr_header__corner__brand chr_brand chr_header__home-menu" title="<?php print htmlspecialchars($site_name); ?>">
           <span class="chr_brand__icon icon-logo"></span>
-        </a>
+          <span>Home</span>
```
AkA84: You might want to pass this string to the `t()` function as well

AkA84:
menus

These menus (`.chr_header__home-sub-menu`, `.chr_header__user-menu__sub-menu`, `.chr_header__settings-sub-menu`) are exactly the same thing in terms of styling, with the only difference being on how they are positioned. 

So I'd strongly suggest to make them one single component (for example: `chr_header__sub-menu`) and let their parents (`.chr_header__home-menu`, `.chr_header__user-menu`, `.chr_header__settings-menu`) dictate their positioning relative to themselves. 

I believe this way you can remove a lot of duplicated and unnecessary css

civihr_default_theme/scss/partials/_menu.scss
```diff
@@ -62,3 +62,102 @@
...
+      color: $light;
+    }
+
+    > i {
+      margin-right: 10px;
+    }
+  }
+}
+
+.chr_header__brand .chr_header__home-sub-menu { //civihr menu in small devices
```
AkA84: Put the comments above the selector, not inline

``` scss
// civihr menu in small devices
.chr_header_brand .chr_header__home-sub-menu {

}
```

civihr_default_theme/templates/page/page.tpl.php
```diff
@@ -56,6 +68,17 @@
       </ul>
     </div>
     <?php } ?>
+    <?php if (user_access("administer users") && user_access("administer nodes") ) { ?>
+    <div class="chr_header__settings-menu">
+      <span class="icon-settings">
```
AkA84: Is this wrapper necessary? Why not simply

``` html
  <i class="fa fa-cog"></i>
  <ul class="chr_header__settings-sub-menu">
  <!-- ... -->
</div>
```

AkA84:
Approved

PR #115, opened 2016-11-18
PCHR-1725: Menu show/hide condition changed

deb1990:
**Requirement**
For civihr_staff or civihr_manager (any role without 'access civicrm' permision):
1. Don't show dropdown on hover
2. Make Home link clickable itself and link to /dashboard' permission):
image-2016-11-16-11-30-56-075

**Fix**
1. Hiding the menu if user does not have `access CiviCRM`
2. Made the whole Home link clickable  if user does not have `access CiviCRM` 

**Update**
The menu items were still visible in the mobile view, as the menu's for Desktop and Mobile view are different. So the following steps were taken to remove them,

1.  `theme_menu_link__dropdown` hook has been implemented to add a class to hide the menu items when access to `civicrm` is not present. 

**Note**
`theme_menu_link__dropdown` is implemented instead of `theme_menu_link` because Radix theme already overrides `theme_menu_link` with `theme_menu_link__dropdown`.


civihr_default_theme/templates/page/page.tpl.php
```diff
@@ -48,16 +48,22 @@
...
-          <span class="chr_brand__icon icon-logo"></span>
-          <span><?php print t("Home"); ?></span>
-        <ul class="chr_header__sub-menu">
-          <?php if (user_access("access CiviCRM")) { ?>
-            <li><?php print $admin_link; ?></li>
-          <?php } ?>
-          <li><?php print $ssp_link; ?></li>
-        </ul>
-      </div>
+      <?php if (!user_access("access CiviCRM")) { ?>
```
AkA84: Couldn't you store this check result in a variable, before all the markup? You are doing the same check three times

``` php
$civi_access = user_access("access CiviCRM");
```

civihr_default_theme/scss/partials/_menu.scss
```diff
@@ -69,8 +69,11 @@
 }
 
 .chr_header__home-menu {
+  color: $light;
```
AkA84, line 72: Are you also adjusting the menu color? Was it part of the ticket requirements? It wasn't mentioned in the PR

civihr_default_theme/templates/page/page.tpl.php
```diff
@@ -48,16 +48,22 @@
...
+        <a href="/dashboard">
+      <?php }?>
+        <div class="chr_header__corner__brand chr_brand chr_header__home-menu" title="<?php print htmlspecialchars($site_name); ?>">
+            <span class="chr_brand__icon icon-logo"></span>
```
AkA84: 4 spaces indentation

civihr_default_theme/templates/page/page.tpl.php
```diff
@@ -48,16 +48,22 @@
...
-        <ul class="chr_header__sub-menu">
-          <?php if (user_access("access CiviCRM")) { ?>
-            <li><?php print $admin_link; ?></li>
-          <?php } ?>
-          <li><?php print $ssp_link; ?></li>
-        </ul>
-      </div>
+      <?php if (!user_access("access CiviCRM")) { ?>
+        <a href="/dashboard">
+      <?php }?>
```
AkA84: I think it's better to use `endif` in these cases, makes the code more readable (http://php.net/manual/en/control-structures.alternative-syntax.php)

AkA84 (changes requested):

civihr_default_theme/scss/partials/_menu.scss
deb1990: Previously the menu was always a `span`, now sometimes its `anchor` when `civicrm` access is not there. So had to explicitly mention the color. Result is same as before.

AkA84 (approved):

civihr_default_theme/scss/partials/_layout.scss
```diff
@@ -79,6 +79,12 @@
     color: $light;
 }
 
+&.not-logged-in {
+    .chr_header__corner {
```
AkA84: You are using 4 spaces indent, instead of 2 spaces (applies to all changes)

civihr_default_theme/template.php
```diff
@@ -355,3 +355,63 @@ class="col-sm-3 control-label ' . ( $label_hidden ? 'hidden-xs' : '' ) . '">'
...
+ * Returns HTML for a menu link and submenu.
+ *
+ * @param $variables
+ *   An associative array containing:
+ *   - element: Structured array data for a menu link.
+ *
+ * @ingroup themeable
+ */
+function civihr_default_theme_menu_link__dropdown($variables)
+{
```
AkA84: The opening parenthesis should be on the same line with `function`

civihr_default_theme/template.php
```diff
@@ -355,3 +355,63 @@ class="col-sm-3 control-label ' . ( $label_hidden ? 'hidden-xs' : '' ) . '">'
   return $fields_structure;
 }
 
+/**
+ * Returns HTML for a menu link and submenu.
+ *
```
AkA84, line 373: You are doing a lot of stuff in this method, please explain what in the documentation

civihr_default_theme/template.php
```diff
@@ -355,3 +355,63 @@ class="col-sm-3 control-label ' . ( $label_hidden ? 'hidden-xs' : '' ) . '">'
...
+    if (in_array('active-trail', $element['#attributes']['class'])) {
+        $element['#attributes']['class'][] = 'active';
+    }
+
+    // Add a unique class using the title.
+    $title = strip_tags($element['#title']);
+    $element['#attributes']['class'][] = 'menu-link-' . drupal_html_class($title);
+
+    /* Code Added */
+    //Hide the links to civicrm and dashboard if Access to CiviCRM is not present
```
AkA84: If we really need to copy paste the entire `menu_link__dropdown` function (and again, I want to make sure that this is the *very* last resort), then at least put your "code added" stuff in a separate function

AkA84 (changes requested):
I'm a bit worried by the fact that the solution consists in copy, pasting and changing the original `menu_link__dropdown` hook contained in radix.

That doesn't look right, and if instead it is, then please provide the documentation/tutorial/something that shows that it is indeed the only proper way to do it

AkA84 (approved):

PR #119, opened 2016-12-06
PCHR-1251: Report tab (static)

AkA84:
This PR contains all style-related changes that were necessary to achieve the design of the report tab, plus generic optimizations

screen

# Technical details
## Colors
There are some new colours added to the theme, as the colours used in the original design do not match any of the colours already listed (`$gray-dark`, `$crm-gray-light-alt`, `$crm-grizzly`, etc).

To keep the scope of the PR small enough, no attempt to normalize the colours list was done. Hopefully over time as new L&A designs are going to be implemented, a pattern will emerge naturally about which colours we should keep and which colour we should discard.

In any case before the L&A extension is shipped, the list of colour should be finalized and the palette used across all pages in the SSP standardized

## Theme structure
Initially the theme structure was as following
```
.
└─ civihr_default_theme/
   └─ scss/
      ├─ base/
      ├─ components/
      ├─ pages/
      ├─ partials/
      ├─ radix/
      └─ civihr_default_theme.style.scss
```
where `partials/` would contain both partials pertaining to drupal components/structure (`_block.scss`, `_page.scss`, `_table.scss`, etc) and to bootstrap components (`_badges.scss`, `_button-groups.scss`, `_tables.scss`, etc)

This obviously would get unmaintainable very soon, so the `scss/partials` folder had been split up into two distinct folders: `scss/drupal` and `scss/partials`, where the former would contains style related to drupal, while the latter would contain only style related to bootstrap
```
.
└─ civihr_default_theme/
   └─ scss/
      ├─ base/
      ├─ components/
      ├─ drupal/
      │  ├─ _block.scss
      │  ├─ _node.scss
      │  ├─ _page.scss
      │  └─ ...
      ├─ pages/
      ├─ partials/
      │  ├─ _buttons.scss
      │  ├─ _panels.scss
      │  ├─ _tables.scss
      │  └─ ...
      ├─ radix/
      └─ civihr_default_theme.style.scss
```

Additionally, in some of the drupal partials there were definitions for custom components (things like `.chr_calendar`, `.chr_header`, etc). Those had been extracted and put in their own separate partials under the `scss/components` folder (`_calendar.scss`, `_header.scss`, etc). Whether we need all those custom components or not is then to be seen, but at least they are placed in the right folder

Finally, all the indentation of the moved partials have been fixed, so that now they have a 2-spaces indentation

## Remove global custom styles
In `base/_custom.scss` there were some global styles for `table` and `body` whose purpose is not really clear and that do not conform with having a Bootstrap framework in place
```scss
body {
  font-family: $crm-sans;
  color: $gray;
  background: $gray-light;
}

.table-responsive {
  margin: 0 -20px;
}
table {
  font-size: 1.3rem;
  width: 100%;
  table-layout: fixed;
}
// etc
```
The body styling was removed and replaced with proper usage of Bootstrap variables
```scss
$body-bg:    $gray-light !default;
$text-color: $gray !default;
$font-family-sans-serif: 'Lato', sans-serif;
```
While the table styling was removed and the bootstrap table partials (`@import "bootstrap/tables";`) in `civihr_default_theme.style.scss` uncommented so that the Bootstrap style for the tables could take over.

**Please note!**
Removing the custom table styling and uncommenting the `bootstrap/tables` partial is sure to have unintented effects on the remaining SSP pages, which relied on those custom styles. In order to keep the scope of this PR to a minimum, for now we ignore these effects, to be addressed later on before the whole L&A extension is shipped

### `.chr_leave-report`
The `.chr_leave-report` component's purpose is to apply all the ad-hoc changes on the report page that are necessary to achieve the look-and-feel of the original design

It cover things like applying the correct color to the action buttons (`.chr_leave-report__action`), giving the desired font weight to the leave requests labels (`.chr_leave-report__label`), setting the width of all the columns of the report table (`.chr_leave-report__table__heading--%{type}`), layout adjustments, etc
### Dropdowns
dropdown

The dropdown style had been slightly changed in terms of borders, text color, highligh background, box shadow

#### `table`, `.table` (amended)
The headings of a table have now a bigger padding than regular cells, a different background color and a darker text color.

table-headings

Also now the first row of a `<tbody>` will have the top border only if there is a heading row preceding it, otherwise it will not have any border

table-no-top-border

#### `.table-condensed` (amended)
A `.table-condensed` has a smaller cell padding than the regular table (but not halved, as in the original Bootstrap settings), *but* it keeps the regular horizontal padding for the leftmost and the outermost cell. Also it removes the background colour for the heading cells
table-condensed

#### `.table-condensed-mixed` (new)
The mixed condensed version keeps the same features of the standard table but the horizontal padding is the same as the one for the condensed version

This variation was introduced so that both the standard table (the outer table in report view) and the condensed tables (the nested tables) can be vertically aligned (see initial screenshot). The variation is generic enough that it can be reused if necessary

#### `.table-nested` (new)
The class is to be applied to the parent element of a `.table` nested within another table
```html
<tr>
  <td class="table-nested-cell">
    <div class="table-nested">
      <table class="table">
```
(notice that the containing `<td>` needs to have the `.table-nested-cell` class applied in order to remove any padding that the cell might have)

When a table has this class applied, it will appear nested via a border effect on the sides, and via a slight inset top & bottom shadow 

nested-table


dhavalsharma:
1. can we move  the values in common elements like `font-weight: 600;` or `line-height: 2.4;` to variables sections?
2. maybe i am missing some build steps but i have run gulp both in theme and in civihr and this the screenshot which I got on chrome 
for admin login
image
and this is how it looks on civihr_staff login 
image

civihr_default_theme/scss/base/_mixins.scss
```diff
@@ -175,3 +175,101 @@
...
+ * Adds a inset shadow to the top/bottom of a nested table
+ *
+ * @param {string} $pos the position of the shadow
+ */
+@mixin table-nested-shadow($pos: 'top') {
+  $pseudo-pos: if($pos == 'top', 'before', 'after');
+  $offset-y: if($pos == 'top', '6', '-4');
+
+  &:#{$pseudo-pos} {
+    box-shadow: inset 0px #{$offset-y}px 9px -6px rgba(0, 0, 0, 0.2);
```
dhavalsharma, line 265: can we use hex colors?

civihr_default_theme/scss/components/_leave-report.scss
```diff
@@ -0,0 +1,56 @@
+.chr_leave-report__action {
+  color: $gray-darker;
+}
+
+
+.chr_leave-report__label {
+  color: $gray-darker;
+  font-weight: 600;
```
dhavalsharma: can 600 be moved to variables?

civihr_default_theme/scss/components/_leave-report.scss
```diff
@@ -0,0 +1,56 @@
...
+.chr_leave-report__label--toggle {
+  cursor: pointer;
+  user-select: none;
+}
+
+
+.chr_leave-report__heading {
+
+  .chr_leave-report__export-button {
+    line-height: 2.4;
```
dhavalsharma, line 24: can 2.4 move to variables?

civihr_default_theme/scss/components/_header.scss
```diff
@@ -0,0 +1,299 @@
...
+
+//CiviHR menu in small devices
+.chr_header__brand .chr_header__home-sub-menu {
+  right: 0;
+  width: 120px;
+  margin: auto;
+}
+
+
+@media (max-width: 767px) {
```
deb1990: Can we add this width in a variable? And keep consistency for all responsive breakpoints.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -0,0 +1,299 @@
...
+  .chr_header__nav__menu > li {
```
deb1990: These selectors can be combined
`.chr_header__nav__menu > li`, `.chr_header__nav__menu > li > a`, `.chr_header__nav__menu > li:last-child > a`
Not sure if this is part of scope of this ticket.

civihr_default_theme/scss/components/_table-w-filters.scss
```diff
@@ -2,88 +2,88 @@
...
-    border-top: 1px solid $gray-light;
+  border-top: 1px solid $gray-light;
 
-    @media(min-width: $screen-md-min) {
-        border-left: 1px solid #ECEDEE;
-        border-top: none;
-        padding-left: 0;
-    }
+  @media(min-width: $screen-md-min) {
+    border-left: 1px solid #ECEDEE;
```
deb1990: This is should be in a variable.

civihr_default_theme/scss/drupal/_calendar.scss
```diff
@@ -0,0 +1,352 @@
...
+  .table-responsive {
+    margin: 0 -20px 0;
+    overflow: auto;
+  }
+
+  @media(min-width: $screen-md-min) {
+    overflow: visible;
+  }
+}
+
```
deb1990: Remove extra lines.

civihr_default_theme/scss/drupal/_form.scss
```diff
@@ -0,0 +1,396 @@
+// Styles for Forms
+// --------------------------------------------------
+
+#civihr-employee-portal-request-new-account-form {
+  label {
+    display: none;
+  }
+}
+
```
deb1990: Remove extra spaces, throughout the file

civihr_default_theme/scss/drupal/_menu.scss
```diff
@@ -0,0 +1,31 @@
...
+// --------------------------------------------------
+
+//Hide Menu link as part of SSP
+.menu-link-home,
+.menu-link-civihr-admin,
+.menu-link-civihr-ssp {
+  display: none;
+}
+
+@media (max-width: 767px) {
```
deb1990: This should go to a variable.

civihr_default_theme/scss/drupal/_view.scss
```diff
@@ -37,37 +37,37 @@
...
+    background: #ffffff;
+  }
 
-    > .view-filters {
-        background: #ffffff;
-        margin-bottom: 20px;
-        padding: $panel-body-padding;
-    }
+  > .view-filters {
+    background: #ffffff;
```
deb1990: Color should be in a variable.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -0,0 +1,299 @@
...
+
+  .chr_header__nav.is-open {
+    padding: 0px
+  }
+}
+
+
+//Hide Menus when not logged in
+.not-logged-in {
+  .chr_header__home-menu .chr_header__sub-menu {
```
deb1990, line 318: This should go to `menu.scss` as we are hiding menus.

deb1990 (changes requested):

civihr_default_theme/scss/components/_header.scss
deb1990: We are already hiding some menu's in menu.scss. This should be clubbed with that code.

deb1990: This one is related to refactoring.

civihr_default_theme/scss/base/_mixins.scss
AkA84: hex colors do not allow you to specify the opacity, hence the usage of `rgba`

civihr_default_theme/scss/components/_leave-report.scss
AkA84: i don't think it makes sense to have it as a variable. It's a specific one-time value used only for this element

deb1990 (approved):

PR #127, opened 2017-01-05
PCHR-1797: Manager Dashboard Styles

deb1990:
This PR contains the styles for the Manager Dashboard page,

**New Components defined**
1. `chr_manager_dashboard` - This is the page component for the Dashboard page
2. `table-clean` - Table with white header and footer
manager_leave___civihr_1_6_demo

3. `chr_vertical_tabs` - Vertical tabs
manager_leave___civihr_1_6_demo

4. `btn-inline` - Button with no top padding which can be used inline

**Partials Modified**
1. Panel -  `panel-subheading` added
2. Table - `tbody` modified to keep same border color for all rows, previously the first row had darker border top.
3. Nav - `active` class styles removed as it was not 

**Styles Removed**
1.`chr_leave-report__action`- Instead `btn-inline` is used.

**Screencast**


civihr_default_theme/scss/components/_button.scss
```diff
@@ -0,0 +1,7 @@
+.chr-button--inline {
```
AkA84: At best this is a variation of the standard `.btn` bootstrap component, so it's definitely not a `.chr_` custom component (notice that you wrote `.chr-`)

But this seems weirdly specific, with also an explicit reference to a dropdown. What is this used for?

civihr_default_theme/scss/components/_table.scss
```diff
@@ -0,0 +1,11 @@
+.chr_table--clean {
```
AkA84: You sure it needs to be a custom component? Why can't it a variation of bootstrap's .table?

civihr_default_theme/scss/components/_manager_dashboard.scss
```diff
@@ -0,0 +1,52 @@
...
+    padding: 0;
+    width: 250px;
+  }
+
+  .panel-title {
+    margin-top: 5px;
+    display: inline-block;
+  }
+
+  > div {
```
dhavalsharma: prefer this to be named by class instead of generic div say chr_manager_dashboard__header_right?

civihr_default_theme/scss/components/_manager_dashboard.scss
```diff
@@ -0,0 +1,52 @@
...
+    font-size: 0.6em;
+    vertical-align: middle;
+    margin-right: 5px;
+  }
+}
+
+.chr_manager_dashboard__sub-header {
+  padding-top: 15px;
+
+  > div {
```
dhavalsharma, line 30: same as above comment , to give it a class name

civihr_default_theme/scss/components/_manager_dashboard.scss
```diff
@@ -0,0 +1,52 @@
...
+    display: inline-block;
+    margin: 0 10px;
+    width: calc(16.4% - 20px);
```
dhavalsharma, line 33: calc is `This is an experimental technology` , is there an alternative for this width?

civihr_default_theme/scss/components/_manager_dashboard.scss
```diff
@@ -0,0 +1,52 @@
...
+    vertical-align: top;
+  }
+}
+
+.chr_manager_dashboard__panel_body {
+  margin: 0;
+  padding: 0;
+
+  [class^="col-"],
+  [class*=" col-"] {
```
dhavalsharma, line 43: can these 2 be combined in `[class~="col-"]` ?

civihr_default_theme/scss/partials/_navs.scss
```diff
@@ -13,11 +13,6 @@
     background-color: transparent;
     color: $brand-primary;
   }
-
-  &.active > a {
-    background-color: $light;
-    color: $gray;
-  }
 }
```
dhavalsharma, line 16: does it have a side affect anywhere else where navs are used?

deb1990: No, I checked.

civihr_default_theme/scss/components/_manager_dashboard.scss
deb1990: CanIUse saying that it is supported in all latest browsers. Also In this case there are no other alternatives in my view.  Also we are using calc in other places for civihr too.

civihr_default_theme/scss/components/_manager_dashboard.scss
deb1990: This is fine in my view, as I want all the child `div`'s of `chr_manager_dashboard__sub-header `to have this style. Adding another class would not have any specific significance.

civihr_default_theme/scss/components/_manager_dashboard.scss
deb1990: To be honest, i am not really sure of the implications of changing this, I read the documentation and it seems that it might work. But Bootstrap too is using like this ( `[class^="col-"],[class*=" col-"]`), so I guess its better to keep like this to keep consistency?

civihr_default_theme/scss/components/_manager_dashboard.scss
deb1990: Updated the Markup so that we dont need to define the class here, instead used bootstrap class `pull-right`

dhavalsharma (approved):

PR #132, opened 2017-01-17
PCHR-1799: Manager Leave Requests Component Styling

deb1990:
As part of this PR(and https://github.com/civicrm/civihr/pull/1346) `ui-select`vendor styles, custom styles for CiviHR has been added.

**Improvements**
Changed Header SubHeader and Vertical tabs background color.


AkA84 (changes requested):
@deb1990 isn't the "multi-select" setting something that you declare on the directive itself (that is, in the view?) The "PCHR-1799: Remove multi select styles" commit has only style changes

deb1990:
@AkA84 Yes multi select is in the directive only. But previously I had set some custom styles for multi-select. I removed that as we are not using multi-select any more.

dhavalsharma (approved):
for anchor tag

PR #137, opened 2017-02-05
PCHR-1273: Calendar Styles

deb1990:
**Changes**

1. Calendar styles moved from HTML to CSS files. The styles were part of the HTML as the plan was to generate Classes dynamically using angular. But it is not the case any more, also the moved classes were meant to be static.

2. Styles for multi select has been added.


civihr_default_theme/scss/components/_ui-select.scss
```diff
@@ -53,6 +53,25 @@
...
+    font-family: "FontAwesome";
+    display: inline-block;
+    height: 33px;
+    line-height: 35px;
+    position: absolute;
+    right: 0;
+    text-align: center;
+    width: 32px;
+    z-index: 1;
+    -webkit-font-smoothing: antialiased;
```
dhavalsharma: font smoothing has limited support [across browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth)? Also only webkit is used here?

dhavalsharma (changes requested):

civihr_default_theme/scss/components/_ui-select.scss
deb1990: I used `font-smoothing` code from `_custom-select.scss` which was already present, now moved the code to a mixin and added `moz` prefix.
Reference: https://davidwalsh.name/font-smoothing

civihr_default_theme/scss/components/_leave-calendar.scss
```diff
@@ -107,7 +107,26 @@ $chr-leave-calendar-legend-size:                  20px;
...
+  border-color: #8ec68a;
+}
+
+.chr_leave-calendar__item--type--weekend {
+  background-color: #c7cbd0 !important;
+  border-color: #c7cbd0 !important;
+}
+
+.chr_leave-calendar__item--type--non-working {
+  background-color: #41afcb !important;
```
AkA84, line 125: Why some have `important!` and some do not?

AkA84 (changes requested):

civihr_default_theme/scss/components/_leave-calendar.scss
deb1990: Kept `!important` for `weekend` and `non-working`day to give them more priority. 
Example: If a date is weekend and also a leaveRequest is there for that date, showing colors for the `weekend` as leave is not calculated for that day.

AkA84 (approved):

dhavalsharma (approved):

PR #144, opened 2017-02-24
PCHR-1421: Style changes for comment

deb1990:
As part of this PR(and https://github.com/civicrm/civihr/pull/1520) comments support has been added to `LeaveRequestModal`


civihr_default_theme/scss/components/_leave-request-modal.scss
```diff
@@ -12,6 +12,10 @@ $row-margin-bottom: 15px;
   overflow: auto;
 }
 
+.chr_leave-request-modal__comments_empty {
+  margin-bottom: 20px;
```
dhavalsharma, line 16: is 20px standard margin bottom, if yes then can it be put in variable?

civihr_default_theme/scss/components/_leave-request-comment.scss
```diff
@@ -1,3 +1,6 @@
+.chr_leave-request-comment{
+  margin-top: 10px;
```
dhavalsharma, line 2: is 10px standard margin top, if yes then can it be put in variable?

dhavalsharma (changes requested):

civihr_default_theme/scss/components/_leave-request-modal.scss
deb1990: At this point we dont have a specific style guide, so everything is mentioned to match the design. Also we have different margin bottom in the same file. 
Its a good point though. Same goes for 2nd review point.

dhavalsharma (approved):

PR #148, opened 2017-03-17
PCHR-1955: Fix logo size and button background in leave email template

mukeshcompucorp:
## Problem 
Different Logo size and "Self Service Portal" button background color in leave email template.

Windows Outlook:
windowsoutlook

## Solution
Add new "soc-td" class and some changes in ".social .soc-btn" class.

## For "Self Service Portal" button:
```
.social .soc-btn {
  padding: 5px 7px 0px 7px;
  color: #FFF;
  font-weight:bold;
  text-align:center;
  display:table-cell;
  width:100%;
  min-width:232px;
}

.soc-td{
  background-color: #DB4A39;
  padding:0 !important;
  vertical-align:middle;
}
```
Windows Outlook:
windowsoutlook


civihr_default_theme/scss/mail.scss
```diff
@@ -55,19 +55,28 @@ table.social {
...
 }
+
 a.fb { background-color: #3B5998!important; }
 a.tw { background-color: #1daced!important; }
 a.gp { background-color: #DB4A39!important; }
 a.ms { background-color: #000!important; }
 
+.soc-td{
+  background-color: #DB4A39;
+  padding:0 !important;
```
AkA84: please always leave a space between the property and the value
```css
padding: 0;
```

AkA84: why the `!important`?

civihr_default_theme/scss/mail.scss
```diff
@@ -55,19 +55,28 @@ table.social {
 
 }
 .social .soc-btn {
```
AkA84, line 57: I understand the need for `.soc-td` (since you're going to wrap the `<a>` into a `<td>`), but what are these changes for `.soc-btn` for?

AkA84 (changes requested):

civihr_default_theme/scss/mail.scss
mukeshcompucorp: Padding: I have changed it to adjust the height of button in all email client except the outlook.
Display: I have changed it because now we are using table structure to wrap `<a>` 
Min-Width: is used for outlook.
Width: is used to adjust the button width for all eamil client except the outlook.

civihr_default_theme/scss/mail.scss
mukeshcompucorp: For Mac Thunderbird we have used important, In manager detail section, Email id is going out side the gray box if we are not using important.

mukeshcompucorp: Fixed.

AkA84 (approved):

PR #151, opened 2017-03-29
PCHR-1862: Fix date picker text getting clipped and toil expiry not aligned

dhavalsharma:
Reduce font size of date text in date picker to prevent its clipping. This is a temporary solution until theme is normalized to adjust date picker text itself.

image

## expiry date for span is aligned correctly
image


AkA84 (changes requested):
can we fix this by using bootstrap grid (`col-sm-6`, etc), instead of ad-hoc magic numbers?

dhavalsharma:
grid ones are not working , either its becoming too long and displacing the other dropdown or too small to clip date

AkA84:
@dhavalsharma can you show a screenshot of that? not sure how it can displaced the other dropdowns if you adjust all the grid cells accordingly

dhavalsharma:
with xs-5 it leaves lots of white space
image

civihr_default_theme/scss/components/_leave-request-modal.scss
```diff
@@ -70,3 +70,16 @@ $row-margin-bottom: 15px;
     }
   }
 }
+
+/*
+* This is a TEMP fix to fit the text of input controls like date picker within input box
+* It is hardcoded to 13px for now.
+*/
+.chr_leave-request-modal__input-font-size {
```
AkA84: this is just a hack, we don't need an entire new selector, can't you just reference classes/tags that are already there? this is a bit overkill

civihr_default_theme/scss/components/_leave-request-modal.scss
```diff
@@ -70,3 +70,16 @@ $row-margin-bottom: 15px;
...
+  font-size: 13px;
+}
+
+.chr_leave-request-modal__span-expiry {
```
AkA84: This looks an ad-hoc fix, did you see what the reason is for the misalingment in the first place

AkA84 (changes requested):

civihr_default_theme/scss/components/_leave-request-modal.scss
dhavalsharma: do you mean something like `chr_leave-calendar__legend__label` but its name suggests its for calendar?

civihr_default_theme/scss/components/_leave-request-modal.scss
dhavalsharma: It is combination of `control-label` and `label` which caused the text to be mis aligned after addition of `uib-datepicker-popup` for expiry date. 
If i use `form-control` in span it gives a border rectangle but aligns it well. is that ok?

AkA84 (approved):
@dhavalsharma approved, but squash the PR review commits

PR #160, opened 2017-05-18
PCHR-2206: Fix contact selection dropdown styles

dhavalsharma:
## Overview
The styles were updated for this [PR](https://github.com/civicrm/civihr/pull/1851).

## Before
The styles of contact were messed up after migration to 1.7.  It looked like below before.

captura de tela de 2017-05-02 08-01-04

## After
This is how it looks with the changes.
eidt-comment

## Technical Details
The styles related to customization of  theme `civihr-ui-select` for dropdown was removed.  This was broken after migration to 1.7. 

The list of contacts were not visible as the `ul` was getting set to 0 opacity, hence changed that. This could be because of `ui-select` usage in a pop up dialog, the drop down list is made invisible. The list is available for user selection but is not visible to the user and hence made it visible.


AkA84:
@dhavalsharma you are missing the "Before" section, also please make sure that jon and mike confirm that the dropdown looks as expected

civihr_default_theme/scss/components/_ui-select.scss
```diff
@@ -74,35 +74,8 @@
 
 /*
 * TEMP fixes till ui-select styling is fixed
+* Dropdown ul opacity was set to 0
```
AkA84, line 77: why? is this an issue with the ui-select library?

civihr_default_theme/scss/components/_ui-select.scss
```diff
@@ -74,35 +74,8 @@
...
 */
-.civihr-ui-select {
```
AkA84, line 78: related to the comment i made in the other pr, why are we removing the entire civihr-ui-select theme? If we had it before, it means it was necessary. What happened that we can just discard it completely? You give no details in the PR description

AkA84 (changes requested):

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: could be , as this is a pop up dialog

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: good questions, tried to to git bisect and changed to earlier version of civicrm, could not locate the cause of regression yet

AkA84 (approved):

PR #161, opened 2017-05-26
PCHR-2208: Fix modal styles

deb1990:
## Overview
Styling issues for the leave request popup has been fixed.

## Before
1. There is some extra space on top of the Tabs which should be removed.
2. Footer section has extra left margin which should be removed
my_leave___hr16-4_

## After
my_leave___hr16-4_


civihr_default_theme/scss/components/_leave-request-modal.scss
```diff
@@ -16,6 +16,10 @@ $row-margin-bottom: 15px;
   margin-bottom: 20px;
 }
 
+.chr_leave-request-modal__footer {
+  margin: 0;
```
AkA84, line 20: why was the margin off in the first place? What caused it? This should be a standard bootstrap modal, the style should work as-is

AkA84 (changes requested):

civihr_default_theme/scss/components/_leave-request-modal.scss
deb1990: The margin was off because `margin-left: -1px` was applied to a very large selector from shoreditch, this probably happened because of using `@include` statement. Spent 1 hour but was not able to figure out. Should I spend more time?

dhavalsharma (approved):

PR #164, opened 2017-05-30
PCHR-2206: Add styles to widen input, border contact dropdown

dhavalsharma:
## Overview
When Manager opened leave request popup the styles of input box were not correct.

## Before
The search input box was getting smaller and the dropdown was losing border for contact selection

## After
This is how dropdown looks now 
image

## Technical Details
Added border to dropdown menu and overridden the width for input dropdown.


civihr_default_theme/scss/components/_ui-select.scss
```diff
@@ -76,6 +76,17 @@
...
 */
-.chr_leave-request-modal__form .ui-select-container ul {
-  opacity: 1 !important;
+.chr_leave-request-modal__form {
+ .ui-select-container ul {
+   opacity: 1 !important;
+  }
+
+  .dropdown-menu {
+    border: 1px solid #ccc;
```
deb1990: Colors should come from already defined variables.

civihr_default_theme/scss/components/_ui-select.scss
```diff
@@ -76,6 +76,17 @@
 * TEMP fixes till ui-select styling is fixed
 * Dropdown ul opacity was set to 0
 */
-.chr_leave-request-modal__form .ui-select-container ul {
-  opacity: 1 !important;
+.chr_leave-request-modal__form {
+ .ui-select-container ul {
```
deb1990, line 80: Leave a line break before every selector.

civihr_default_theme/scss/components/_ui-select.scss
```diff
@@ -76,6 +76,17 @@
...
+   opacity: 1 !important;
+  }
+
+  .dropdown-menu {
+    border: 1px solid #ccc;
+  }
+
+  .ui-select-container:not(.ui-select-multiple) input {
+    /*the width was getting set to 503px in element itself thus causing the input box to reduce*/
+    width: 100% !important;
```
deb1990, line 91: In manager leave page's Filter section and in calendar, UI Select is used. They are not broken by this change right?

deb1990 (changes requested):

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: there is no color for #ccc and this is its only usage so not added

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: this is inside popup class 
```
.chr_leave-request-modal__form {
 .ui-select-container ul {
   opacity: 1 !important;
  }

  .dropdown-menu {
    border: 1px solid #ccc;
  }

  .ui-select-container:not(.ui-select-multiple) input {
    /*the width was getting set to 503px in element itself thus causing the input box to reduce*/
    width: 100% !important;
  }
}
```

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: as you can see in all other selectors the first selector is below parent and rest of them have line breaks like 

```
.ui-select-bootstrap .ui-select-choices-row {
  > span {
```
```
.ui-select-container {
  &:not(.ui-select-multiple) {
```
this is just consistent with them

civihr_default_theme/scss/components/_ui-select.scss
```diff
@@ -76,6 +76,17 @@
 * TEMP fixes till ui-select styling is fixed
 * Dropdown ul opacity was set to 0
 */
-.chr_leave-request-modal__form .ui-select-container ul {
-  opacity: 1 !important;
+.chr_leave-request-modal__form {
```
deb1990, line 79: Sorry did not notice this before, but `_ui-select.scss`should contain only the global styles. This should go to `_leave-request-modal.scss`

civihr_default_theme/scss/components/_ui-select.scss
deb1990: It was wrong then, please fix it for all scenarios of this file.

civihr_default_theme/scss/components/_ui-select.scss
deb1990: First of all I think adding new color is not necessary and we can use the same color of the input box.(Pending confirmation from @AkA84 ) If we decide to add new color, it should be defined in the variables. No colors should be used without variables, that is what we are following in civihr. Also you can search for a already defined color which is close to `#ccc` and use the same.

deb1990 (changes requested):

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: i will be consistent with rest of the file here and are we following that in test files where there is no newline for the first describe

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: this is only a temp fix and will go later

civihr_default_theme/scss/components/_ui-select.scss
dhavalsharma: as discussed in lna room, updated accordingly

deb1990 (approved):

PR #173, opened 2017-06-16
PCHR-2035: Variables used for select dropdown

deb1990:
## Overview
Variables are used in shoreditch(https://github.com/compucorp/org.civicrm.shoreditch/pull/19) for `height` and `border-color`, so this repository need to be updated as it imports partials from Shoreditch.

## Technical Details
`$input-border` variable set to `$crm-color-neutral` as it is the default border color is this theme.
Also `$crm-color-neutral` is replaced by `$input-border` wherever it was used for border.

## Comment
Can only be merged once, https://github.com/compucorp/org.civicrm.shoreditch/pull/19 is merged and 1.7 is synced.


civihr_default_theme/scss/base/_variables.scss
```diff
@@ -163,3 +163,8 @@ $crm-scrollbar-thumb-color: $crm-gray-light-blue !default;
 $crm-customselect-height:        33 !default;
 $crm-customselect-caret-width:   32 !default;
 $crm-customselect-padding-right: $crm-customselect-caret-width + 12 !default;
+
+//== Forms
+//
+//##
+$input-border:              $crm-color-neutral !default;
```
AkA84, line 170: is `$input-border` a default bootstrap SASS variable? If yes, what impact does this have on the SSP theme as a whole?

AkA84 (changes requested):

civihr_default_theme/scss/base/_variables.scss
deb1990: Yes it is bootstrap, other components were using `$crm-color-neutral` for the border, so replaced it with `$input-border`. Added in PR document too.

AkA84: so you defined the value of a default bootstrap variable, shouldn't have an effect on the entire theme? (i have to assume that `$input-border`, being a default variable, is used across several SCSS partials)

deb1990: I have checked that `$input-border`is not used anywhere else other than the `form-control`, also `$crm-color-neutral` was used previously for defining the border which was wrong. So using `input-border` makes sense as it internally is same as `$crm-color-neutral`.

AkA84 (approved):

PR #182, opened 2017-07-04
PCHR-2033: Move styles to L&A

deb1990:
`_manager_calendar.scss` is moved to Leave and Absence extension(https://github.com/civicrm/civihr/pull/1981).


civihr_default_theme/scss/civihr_default_theme.style.scss
```diff
@@ -117,6 +117,7 @@
 @import "SCSSROOT/uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/tables";
 @import "SCSSROOT/uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/leave-report";
 @import "SCSSROOT/uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/leave-calendar";
+@import "SCSSROOT/uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/manager_calendar";
```
igorpavlov-zz, line 120: I am not sure this should be called manager calendar. We might consider renaming files as it becomes a shared component.

igorpavlov-zz (changes requested):

civihr_default_theme/scss/civihr_default_theme.style.scss
deb1990: I think `manager_calendar` is still fine because still it shows the calendar for the managees. Admin is getting all contacts because he manages all.

igorpavlov-zz (approved):

PR #191, opened 2017-07-18
PCHR-2417: Fix table styles

deb1990:
## Overview
Table cell styles on 1.7 SSP dashboard are different from 1.6, this PR fixes that.

## Analysis
As part of https://github.com/compucorp/civihr-employee-portal-theme/commit/f23f3588eeb409aa2f9baf25b490eb75885372f6#diff-408fa5bab1ade4518c7b975a457a71b3 styles for default tables were removed from `custom.scss` and `@import "bootstrap/tables"` was enabled but this change went to 1.7 and not 1.6 hence the difference in look.

## Technical Details
1. Default table styles removed from `custom.scss`(same as 1.7)
2. `@import "bootstrap/tables"` enabled (same as 1.7)
3. Contents of `partials/_table.scss` has been moved to `layout/_table.scss` because it contains styles for specific pages.
4. Custom styles for default table is added to `partials/_table.scss`.
5. Added `civihr_default_theme_preprocess_views_view_table` and `civihr_default_theme_table` hook to remove the `table-striped` and `table-bordered` classes from default table. This is because radix theme always adds these 2 classes which is not required in our theme.

1.6 SSP
dashboard___hr16-7

1.7 SSP
dashboard___hr17-1

## Comment
1.7 need to be synced with 1.6 after merging this PR.

---

- [x] Tests Pass


civihr_default_theme/scss/layout/_table.scss
```diff
@@ -0,0 +1,180 @@
+/**
+ * Manager documents table
+ *
+ * # TEMPORARY #
+ */
+#documents-dashboard-table-manager {
+    table-layout: auto;
```
AkA84: fix the indentation

civihr_default_theme/template.php
```diff
@@ -14,6 +14,169 @@
...
+ * Implements template_preprocess_views_view_table().
+ */
+function civihr_default_theme_preprocess_views_view_table(&$variables) {
+  $variables['classes_array'] = array_diff($variables['classes_array'], array('table-striped', 'table-bordered'));
+}
+
+/**
+ * Implements theme_table().
+ */
+function civihr_default_theme_table($variables) {
```
AkA84: this code looks *massive*, did you copy it from somewhere else? Please split it into smaller functions, then i'll re-review

AkA84 (changes requested):
@deb1990 since you created backstopjs tests, please show the before/after diffs

civihr_default_theme/template.php
deb1990: Yes it was copied from radix. I have added a comment specifying which part has been changed. Otherwise I dont think its a good idea to split the functions as its copied from radix, so incase radix function changes we can easily copy that back to here .

AkA84 (approved):

PR #214, opened 2017-08-31
PCHR-2526: Add style for the use yoti button

bmnepali:
## Overview
Customize the "Use yoti" button for small font and center alignment.

## Before
1. Alignment of the button is not centered and font is small
screenshot-2

## After
2. Center aligned the button with small font size.
screen shot 2017-08-31 at 3 24 38 pm

## Technical Details
1. Add styles for partials/_block.scss partial  in "civihr_employee_portal_theme/civihr_default_theme" for "Use yoti" button
```css
// Use yoti button
.yoti-connect {
  text-align: center;

  a {
    font-size: 14px !important;
  }
}
```


civihr_default_theme/scss/partials/_block.scss
```diff
@@ -75,4 +75,13 @@
   .form-submit {
     min-width: 82px;
   }
-}
\ No newline at end of file
+}
+
+// Use yoti button
+.yoti-connect {
```
mickadoo, line 81: Is the class still 'yoti-connect' even after the extension was renamed to 'yoti'?

bmnepali: Yes! the class still is `yoti-connect`.
screen shot 2017-08-31 at 3 59 14 pm

mickadoo: Ah, you're right, looks like this is the [code](https://github.com/getyoti/yoti-drupal-7/blob/master/yoti/yoti.module#L150). Confusing they still use the old name in places though..

civihr_default_theme/scss/partials/_block.scss
```diff
@@ -75,4 +75,13 @@
...
+  text-align: center;
+
+  a {
+    font-size: 14px !important;
```
mickadoo, line 85: I'm no FE dev, but I know important should be used sparingly, is it possible to get the same result with a more specific CSS selector? If not then this is fine.

mickadoo (approved):
One question left about CSS selector, but it's not really my field so I'll leave it up to you.

civihr_default_theme/scss/partials/_block.scss
bmnepali: Well, you right! And its mandatory to use `!important` here as the font size is defined as default `inline` in the button markup and we need to customize the default font size. Since the inline style has the higher specificity, we must use `!important` to override it form css.

PR #222, opened 2017-09-12
PCHR-2573: Fix table margin inside panel

igorpavlov-zz:
## Overview

This PR simply fixes the tables' margin that appear inside `.panel` element as per the Styleguide.

## Before

n

## After

a

## Technical Details

An extra rule is added:

```css
.panel table, .table-responsive {
  margin-bottom: 0;
}
```


reneolivo:
This is a very small change, but can we change how it affects other areas?

igorpavlov-zz:
@reneolivo what I've done in the PR Feedback is I removed all "local" fixes for tables margins and applied a new rule `PANEL->TABLE=MARGIN-BOTTOM:0` in Shoreditch and SSP Theme. Scanned through all components and it looks great. Not only that, the mobile view is now also fixed.

reneolivo (approved):

PR #232, opened 2017-09-20
PCHR-2508: Menu Responsivenes Fix

deb1990:
## Overview
As part of this PR, the responsiveness of the SSP Menu has been improved. 
The gear icon and the user name was getting shifted to the next line, it has been fixed.

## Before
resize

## After
2508a

## Technical Details
1. A new variable `$crm-menu-responsive-breakpoint` has been created. The same is used as the media query variable for `.chr_header__user-menu__name`.

---

- [x] Tests Pass


civihr_default_theme/scss/components/_header.scss
```diff
@@ -1,3 +1,5 @@
+$chr-menu-responsive-breakpoint: 1250px;
```
AkA84: This should go in the base/_variables.scss. Also it seems that the prefix we are using here is for the variables is `crm_` (this i think is my fault, back in the day when i just joined), so let's try to keep things consistent

AkA84 (changes requested):

civihr_default_theme/scss/base/_variables.scss
```diff
@@ -116,7 +116,8 @@ $nav-tabs-justified-active-link-border-color: $light !default;
 // Media queries breakpoints
 // ------------------------------------------------------------------------------------------------------
 
-$screen-sm-max:               767px !default;
+$screen-sm-max:                   767px !default;
+$crm-menu-responsive-breakpoint:  1250px;
```
AkA84, line 120: It's `crm_`, not `crm-`

AkA84 (approved):
i need to leave for 20 mins or so, so i'm going to approve anyway, but there is a small fix to do

civihr_default_theme/scss/base/_variables.scss
deb1990: Its `crm-` only, I have checked other instances, 
For example: https://github.com/compucorp/civihr-employee-portal-theme/blob/d0c40f87036964df1f9b2d013de26b05415c46c0/civihr_default_theme/scss/base/_variables.scss#L126-L126

deb1990:
@AkA84 I am merging the ticket

PR #248, opened 2017-10-18
PCHR-2706: Profile Pic Style

deb1990:
## Overview
As part of this PR, the Profile Picture page of Onboarding Wizard has been styled.

## Technical Details
1. `Drupal.civihr_theme.createDragAndDrop` has been added. It accepts the css selector for the input field, and gives it a Drag and Drop Uploader style, as shown in the screenshot.
---


civihr_default_theme/scss/pages/_onboarding-wizard.scss
```diff
@@ -195,4 +214,50 @@
...
+    display: flex;
+    justify-content: center;
+    height: 180px;
+    outline: 1px dashed #7a8084;
+    pointer-events: none;
+    position: absolute;
+    text-align: center;
+    transition: outline-offset .15s ease-in-out, background-color .15s linear;
+    top: 0;
+    width: 475px;
```
reneolivo: If this theme supports mobile devices, this should be something like:

```scss
width: 100%; // which is the default value, so this is not needed
min-width: 120px; // an arbitrary number. Any number that allows the input to look well even in small screens and containers.
```

This applies to the input width above too.

reneolivo (changes requested):

reneolivo (approved):

PR #251, opened 2017-10-24
PCHR-2709: Password, Customize Welcome Wizard, Responsive Styles 

deb1990:
## Overview
As part of this PR the following things are done
- Password Page Styles
- Customize Welcome Wizard Page Styles
- Responsive Styles for Progress bar, Form Pages.

## How it Looks
**Password Page**
localhost-8910-user-1-edit

**Customize Welcome Wizard Page**
localhost-8910-customize-onboarding-wizard

**Responsiveness**
localhost-8910-user-1-edit iphone 6 plus

localhost-8910-onboarding-form iphone 6 plus 1

## Technical Details

---

- [ ] Tests Pass


civihr_default_theme/scss/pages/_onboarding-wizard.scss
```diff
@@ -377,4 +463,154 @@ $carousel-box-shadow-color: rgba(#3f3c55,0.3);
...
+      width: 440px;
+      position: absolute;
+      top: 850px;
+      right: 100px;
+    }
+
+    .onboarding-customize-welcome {
+      width: 440px;
+      position: absolute;
+      top: 1200px;
```
reneolivo, line 612: Is there a reason why these elements are being positioned absolute instead of following a natural flow (ie, only using margin and padding)?

What would happen if the customize features content expands? would it overlap with the customize welcome section?

reneolivo (approved):

civihr_default_theme/scss/pages/_onboarding-wizard.scss
deb1990: Unfortunately this was the only way I found to make this work, because the images are injected in the page using JS, and they do not belong to the original flow of the HTML.

PR #254, opened 2017-10-25
PCHR-2631: Update cog menu

AkA84:
## Overview
This PR amends the "cog icon" menu in the SSP top menu
 
## Before
before

## After
after

## Technical Details
A bit of refactoring was done, to remove the logic of building the menu from the page template to the `template.php` file

PR #255, opened 2017-10-26
PCHR-2629: Update menu style

AkA84:
## Overview
This PR introduces the new style of the SSP main menu

## Before
menu-before

## After
menu-after

### Implemented `.chr_logo`
The new [logo component](https://github.com/civicrm/civihr/blob/PCHR-2578-menu-updates/org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_logo.scss) introduced here #2236 had been used instead of the old logo

You can see the differences between the two logos in the header in the screenshot above, for the footer please check these two screenshots instead

*Before*
footer-logo-before

*After*
footer-logo-after

### Fixed visibility of the cog menu item
The cog menu item, introduced [here](https://github.com/compucorp/civihr-employee-portal-theme/pull/254) was being displayed also on the welcome page, when the user was logged in

cog-logged-out-before

This was due to the separator being always added to the markup, even if the user, when not logged in, obviously didn't have any permission to see the actual menu items
```php
function _build_cog_menu_markup() {
  foreach($menuItems as $menuItem) {
    if (isset($menuItem['separator']) && $menuItem['separator'] == TRUE) {
      $markup .= '<hr class="chr_header__sub-menu__separator">';
      continue;
    }
  }
}
```
This led to the markup to never being empty, and thus the icon was always visible
```php
<?php if (!empty($cog_menu_markup)) { ?>
      <i class="fa fa-cog" aria-hidden="true"></i>
      <ul class="chr_header__sub-menu">
        <?php print $cog_menu_markup; ?>
      </ul>
    </div>
<?php } ?>
```
A simple check on whether the user is logged in or not before even building the markup fixed the problem
```php
$variables['cog_menu_markup'] = $variables['logged_in'] ? _build_cog_menu_markup() : NULL;
```
cog-logged-out-after


civihr_default_theme/scss/components/_header.scss
```diff
@@ -150,14 +161,17 @@
 
   @media(min-width: $screen-md-max) {
     a {
-      color: inherit;
+      color: $light-pale;
```
reneolivo, line 164: Do we have colours for different screen sizes? Why does this have to be placed inside a `@media` block?

civihr_default_theme/scss/components/_header.scss
```diff
@@ -150,14 +161,17 @@
...
       padding: 0 15px;
 
-      &.active,
       &:hover,
       &:focus {
         background: $gray;
-        color: $light;
+        color: $light !important;
```
reneolivo, line 170: Same question as before

civihr_default_theme/scss/components/_header.scss
AkA84: Mobile menu

mobile-menu

civihr_default_theme/scss/components/_header.scss
AkA84: See above

reneolivo (approved):

PR #260, opened 2017-10-29
PCHR-2639: Restrict access to hr admin link

AkA84:
## Overview
This PR makes the HR admin link visible only to users with the "administer CiviCRM" permission

### Cog menu item
The way the item was built had been refactored. Before, the separator was an entire separate item that the `_build_cog_menu_markup` function would then add to the markup
```js
[
  [
    'permissions' => ["access content overview"],
    'link' => l(t('Manage HR Resources'), 'admin/content', array('html' => TRUE)),
  ],
  [
    'permissions' => ["edit terms in {$resourceTypeVocabularyID}"],
    'link' => l(t('HR Resource Types'), 'hr-resource-types-list', array('html' => TRUE))
  ],
  [
    'separator' => TRUE
  ],
  [
    'permissions' => ["administer users", "access users overview"],
    'link' => l(t('Manage Users'), 'admin/people', array('html' => TRUE)),
  ],
];
```
```js
function _build_cog_menu_markup() {
  if (isset($menuItem['separator']) && $menuItem['separator'] == TRUE) {
    $markup .= '<hr class="chr_header__sub-menu__separator">';
    continue;
  }
}
```
This turned out to be problematic for when a user did not have any permission to see the actual submenu items. Even in that case the menu was never really empty as it would be expected to be, because of the separator item being always added to the markup. This caused the user to see the cog icon even when there was no submenu to display
```php
<?php if (!empty($cog_menu_markup)) { // there are no real items in here, only the separator?>
  <i class="fa fa-cog" aria-hidden="true"></i> <!-- the icon gets printed anyway -->
  <ul class="chr_header__sub-menu">
    <?php print $cog_menu_markup; ?> 
  </ul>
</div>
<?php } ?>
```
The solution to this was to make the separator an attribute (that gives the item a special styling) instead of an item
```php
[
  [
    'permissions' => ["access content overview"],
    'link' => l(t('Manage HR Resources'), 'admin/content', array('html' => TRUE)),
  ],
  [
    'permissions' => ["edit terms in {$resourceTypeVocabularyID}"],
    'link' => l(t('HR Resource Types'), 'hr-resource-types-list', array('html' => TRUE)),
    'separator' => TRUE,
  ],
  [
    'permissions' => ["administer users", "access users overview"],
    'link' => l(t('Manage Users'), 'admin/people', array('html' => TRUE)),
  ],
];
```
This way only the real items would be added to the markup
```php
function _build_cog_menu_markup() {
  foreach($menuItems as $menuItem) {
    foreach ($menuItem['permissions'] as $permission) {
      if (user_access($permission)) {
        $markup .= _get_cog_menu_item_markup($menuItem);
        break;
      }
    }
  }
}

function _get_cog_menu_item_markup($menuItem) {
  $class = '';

  if (isset($menuItem['separator']) && $menuItem['separator'] == TRUE) {
    $class = 'chr_header__sub-menu__separator';
  }

  return "<li class=\"{$class}\">{$menuItem['link']}</li>";
}
```
and as such if the user could not see any of the items, the markup would then truly empty


civihr_default_theme/template.php
```diff
@@ -413,6 +406,23 @@ function _get_cog_menu_items() {
 }
 
 /**
+ * Builds and returns the markup of the given menu item
+ *
+ * @param array $menuItem
+ *
```
deb1990, line 412: There should not be a line break here.

deb1990 (changes requested):

deb1990:
@AkA84 I have reviewed but as it is mostly about PHP changes, I would suggest someone from backend to review it once. I have added 1 comment.

civihr_default_theme/template.php
AkA84: In PHP you need to leave new lines between `@params` and `@return`

civihr_default_theme/template.php
```diff
@@ -400,10 +395,8 @@ function _get_cog_menu_items() {
     ],
     [
       'permissions' => ["edit terms in {$resourceTypeVocabularyID}"],
-      'link' => l(t('HR Resource Types'), 'hr-resource-types-list', array('html' => TRUE))
-    ],
-    [
-      'separator' => TRUE
+      'link' => l(t('HR Resource Types'), 'hr-resource-types-list', array('html' => TRUE)),
```
davialexandre: Please use the short array syntax to last `l()` param

civihr_default_theme/template.php
```diff
@@ -660,61 +660,97 @@ function _hide_menu_items(&$element) {
...
-  // Fix for active class.
-  if (($element['#href'] == current_path() || ($element['#href'] == '<front>' && drupal_is_front_page())) && (empty($element['#localized_options']['language']) || $element['#localized_options']['language']->language == $language_url->language)) {
-    $element['#attributes']['class'][] = 'active';
+/**
+ * If the link is for the current page, then it adds the active class to it
+ *
+ * @param array $link
+ */
+function _add_active_class_if_link_is_current(&$link) {
+  if (($link['#href'] == current_path() || ($link['#href'] == '<front>' && drupal_is_front_page())) && (empty($link['#localized_options']['language']) || $link['#localized_options']['language']->language == $language_url->language)) {
```
davialexandre: Quite a complex condition here. Please split it into smaller bits to make it easier to read. Something like:

```php
$linkToFrontPage = $link['#href'] == '<front>' && drupal_is_front_page();
$linkToCurrentPage = $link['#href'] == current_path();
...
if($linkToCurrentPage || $linkToFrontPage) {
...
}
```

Also, can you please explain what do you need the localized_options check for?

civihr_default_theme/template.php
```diff
@@ -660,61 +660,97 @@ function _hide_menu_items(&$element) {
...
+
+  return '<li' . drupal_attributes($element['#attributes']) . '>' . _get_link_markup($element) . "</li>\n";
+}
 
-  // Fix for active class.
-  if (($element['#href'] == current_path() || ($element['#href'] == '<front>' && drupal_is_front_page())) && (empty($element['#localized_options']['language']) || $element['#localized_options']['language']->language == $language_url->language)) {
-    $element['#attributes']['class'][] = 'active';
+/**
+ * If the link is for the current page, then it adds the active class to it
+ *
```
davialexandre: Looking at the function implementation, it seems that are many conditions for a page to be considered the current one. Perhaps it would be better to augment the function's description with more details about what is the current page

civihr_default_theme/template.php
```diff
@@ -660,61 +660,97 @@ function _hide_menu_items(&$element) {
...
-  } else {
-    $output = l($element['#title'], $element['#href'], $element['#localized_options']);
+/**
+ * Hide the link to the admin if the current user does not
+ * have administer access
+ *
+ * @param array $link
+ */
+function _hide_admin_link_to_basic_users(&$link) {
+  $adminAccess = user_access("administer CiviCRM");
```
davialexandre: Please use single quotes here

civihr_default_theme/template.php
```diff
@@ -660,61 +660,97 @@ function _hide_menu_items(&$element) {
...
 
-  return '<li' . drupal_attributes($element['#attributes']) . '>' . $output . $sub_menu . "</li>\n";
+/**
+ * Build and returns the markup of the given link (and submenu, if present)
+ *
+ * @param array $link
+ *
+ * @return string
+ */
+function _get_link_markup(&$link) {
```
davialexandre: It looks like this function only works for menu links, so I think it would be better to give it a more specific name, like `_get_link_markup()`

davialexandre (changes requested):

civihr_default_theme/template.php
AkA84: > Also, can you please explain what do you need the localized_options check for?

This is from [radix](https://www.drupal.org/project/radix). The whole `civihr_default_theme_menu_link__dropdown` function is an override of the radix's one (basically, a copy paste with customizations on top)

davialexandre (approved):

PR #261, opened 2017-10-30
PCHR-2112: Bug fixes

deb1990:
## Overview
This PR fixes bugs related to the Onboarding Wizard

## Issue 1
**Problem**
The DOB format is different as on laptop. Android: DD/MM/YYYY, Laptop: YYYY-MM-DD

**Solution**
The mobile version format cannot be changed, so changed the desktop format to match the Mobile version.
welcome_to_civihr___hr17-4

## Issue 2
**Problem**
In Iphone the DOB field's design seems broken on Step2 page.

iphone_x_-_ios_11_0_and_pchr-2112__bug_fixes_by_deb1990_ _pull_request__261_ _compucorp_civihr-employee-portal-theme

## Issue 3
**Problem**
The progress bar is not responsive enough. It breaks on 13 inch laptops

responsive

## Issue 4
**Problem**
Different UI colors were wrong

welcome_to_civihr___hr17-4

## Issue 5
**Problem**
Calendar was broken on non onboarding pages

**Solution**
Previously the webform-calendar template was overridden which changed the fields behaviour for all pages. Changed the same to override only for the Onboarding Pages.

## Others
Refactored the `drop-layer` class into more meaningful `drop-zone-overlay` name, and moved the styles into a separate file.


civihr_default_theme/assets/js/civihr_default_theme.script.js
```diff
@@ -234,10 +239,22 @@
    * Set values to Webform's SELECT input type calendar values from
    * Native Datepicker
    */
-  function setWebformCalendarValues () {
-    var date = new Date(this.value);
-    $('#' + this.id + '-' + 'month').val(date.getMonth() + 1);
-    $('#' + this.id + '-' + 'day').val(date.getDate());
-    $('#' + this.id + '-' + 'year').val(date.getFullYear());
+  function setWebformCalendarValues (dateText) {
```
reneolivo, line 290: please document `dateText` param.

civihr_default_theme/assets/js/civihr_default_theme.script.js
```diff
@@ -234,10 +239,22 @@
...
+    var month, date, year, fullDate;
+
+    if (Drupal.behaviors.isMobile) {
```
reneolivo: can we change this part to something like:

```js
var dateValues = Drudal.behaviors.isMobile
  ? getMobileCalendarValues()
  : getDesktopCalendarValues() // note: you may need to pass the right params

$('#' + this.id + '-' + 'month').val(dateValues.month).trigger('change');
// etc
```

civihr_default_theme/scss/base/_variables.scss
```diff
@@ -118,6 +118,7 @@ $nav-tabs-justified-active-link-border-color: $light !default;
 
 $screen-sm-max:                   767px !default;
 $crm-menu-responsive-breakpoint:  1250px;
+$crm-onboarding-wizard-menu-responsive-breakpoint:  1300px;
```
reneolivo, line 121: if this variable is used only for the onboarding wizard, why define it in the base variables file?

civihr_default_theme/scss/pages/_onboarding-wizard.scss
```diff
@@ -79,6 +79,25 @@ $carousel-box-shadow-color: rgba(#3f3c55,0.3);
       li {
         display: inline-block;
         float: none;
+
+        a {
+          color: #4c4c5d !important;
```
reneolivo: $gray-darker

reneolivo (changes requested):

civihr_default_theme/scss/base/_variables.scss
deb1990: Here we have all the responsive variables defined, just like `$crm-menu-responsive-breakpoint`. So kept all at the same place.

reneolivo (approved):

PR #265, opened 2017-11-03
PCHR-2525: Remove Unwanted Styles

deb1990:
## Overview
As part of this PR, styles related to `ui-select` has been removed, as it is taken care from shoreditch(https://github.com/compucorp/org.civicrm.shoreditch/pull/65).


civihr_default_theme/scss/partials/_forms.scss
```diff
@@ -30,17 +30,6 @@
 // Theme has issues with ui-select search field
```
AkA84: shouldn't you remove this as well?

civihr_default_theme/scss/partials/_forms.scss
```diff
@@ -30,17 +30,6 @@
...
-}
-
-.ui-select-multiple {
-
-  .ui-select-search {
-    width: 150px !important;
-  }
-}
-
 .ui-select-toggle {
```
AkA84: what about this?

AkA84 (changes requested):

civihr_default_theme/scss/partials/_forms.scss
deb1990: Yes it is also not required, the default value of `ui-select-toggle` is also `7px 10px`is SSP.

AkA84 (approved):

PR #269, opened 2017-11-07
PCHR-2881: Remove hover effect on CiviHR logo + refactoring

AkA84:
## Overview
The CiviHR logo (visible on medium to small viewports) used to trigger the mobile menu, but doesn't anymore since #258 , but the hover effect was kept. This PR removes the hover effect

## Before
hover

## After
no-hover

## Technical Details
This line
```php
$identifier = $link['#localized_options']['identifier'];
```
would cause a notice message when the property `identifier` would not exist in the `#localized_options` array

A safer approach has been adopted to avoid the message
```php
$isAdminLink = isset($localOptions['identifier']) && $localOptions['identifier'] == 'main-menu_civihr-admin:civicrm';
```

PR #275, opened 2017-11-13
PCHR-2920: Fix logo hover color on desktop

AkA84:
## Overview
#269 removed the hover color from the logo that appears on medium-to-small viewports (since it wasn't a link anymore) but accidentally removed it also from the logo that appears on large viewports (which is instead an actual link). This PR fixes the issue

## Before
before

## After
after


civihr_default_theme/scss/components/_header.scss
```diff
@@ -87,6 +87,10 @@
     display: none;
     height: 50px;
     text-align: center;
+
+    &:hover {
+      color: rgba($light, 0.8);
```
deb1990, line 92: This will affect mobile devices also I think. In mobile you dont have any hover event, but the when you tap the element, the hover css gets applied. Could you check if that is the case here.

deb1990 (changes requested):

civihr_default_theme/scss/components/_header.scss
AkA84: @deb1990 it doesn't, as it targets the logo in the "corner" element, which is shown only on desktop
```scss
.chr_header__corner {
  > .chr_header_logo {
    &:hover {
    }
  }
}

deb1990 (approved):

PR #278, opened 2017-11-22
PCHR-2947: Style HR Details

deb1990:
## Overview
This PR(and https://github.com/compucorp/civihr-employee-portal/pull/399) styles the newly created Emergency and Dependents blocks and their modal popup in HR Details Tab.

## Before

bef

## After

**Desktop**

**iPad**
ipad

**iPhone**
iphone

## Technical Details
1. Styles related to HR Details page, has been moved to the newly created `_hr-details.scss` page.
2. Removed padding and margin from `.modal-dialog` for mobile devices, so that the popup takes up the full page. Refer `iPhone` screenshot. This change is for all modal popup in SSP.

---


igorpavlov-zz:
@deb1990 could you please include some gifs/screenshots on how it was before?

igorpavlov-zz:
@deb1990 please specify which exact tests are passed

deb1990:
@igorpavlov This PR styled the Emergency and Dependents blocks which are newly created, still added a GIF.
And for this PR, tests are not necessary, so removed that part.

civihr_default_theme/scss/pages/_hr-details.scss
```diff
@@ -0,0 +1,63 @@
+/* Page: My details */
+
+.page-hr-details {
+  .civihr-employee-portal-my-details-block {
+    background: $light;
+    padding: 20px;
+    margin: 0 15px;
```
igorpavlov-zz, line 7: There are a lot of places in this change that mention 15px, I wonder if they are connected. If they are, then 15px should go to variables. Moreover, I think the name of the variable should *not* be something like "block-margin", but rather smth like "block-gap" so this var can be reused in both margins, paddings etc.

igorpavlov-zz (changes requested):

igorpavlov-zz (approved):

civihr_default_theme/scss/pages/_hr-details.scss
deb1990: I looked at all the scss from SSP theme, looks like we have a mixture of 15px and 20px all over the place. 
So 
1. Creating a single variable would not solve this problem as we have different values
2. We could try to use a single value like either 15 or 20, but then the change will be huge and scope will increase a lot. And this ticket needs to be done by today. cc @mi-kon

igorpavlov-zz: OK, but please consider refactoring it when there is a chance since having these values all over the place can be an issue for a further development.

PR #294, opened 2018-01-15
PCHR-3170: Re-add "Customize Welcome Wizard" Link

mickadoo:
## Overview

When merging the changes from menu refactoring into the onboarding branch the cog menu item for "Customize Welcome Wizard" was dropped in a [merge conflict resolution](https://github.com/compucorp/civihr-employee-portal-theme/commit/b5669798a1216615f7ce4d637b7ebefa17995fad#diff-4b16c32fda3f6db9862b392f76668cdeL23). This needs to be re-added.

## Before

The "Customize Welcome Wizard" cog menu item is missing

image

## After

The "Customize Welcome Wizard" cog menu item is back

image

## Comments

This repository did not have any PHP tests until now, but this PR adds the first PHPUnit test to check that the expected link markup will be returned.

---

- [x] Tests Pass

### DIFF
--- a/civihr_default_theme/template.php
+++ b/civihr_default_theme/template.php
@@ -494,19 +494,25 @@ function _build_cog_menu_markup() {
 function _get_cog_menu_items() {
   $resourceTypeVocabularyID = taxonomy_vocabulary_machine_name_load('hr_resource_type')->vid;
 
+  $options = ['html' => TRUE];
+
   return [
     [
       'permissions' => ["access content overview"],
-      'link' => l(t('Manage HR Resources'), 'admin/content', ['html' => TRUE]),
+      'link' => l(t('Manage HR Resources'), 'admin/content', $options),
     ],
     [
       'permissions' => ["edit terms in {$resourceTypeVocabularyID}"],
-      'link' => l(t('HR Resource Types'), 'hr-resource-types-list', ['html' => TRUE]),
+      'link' => l(t('HR Resource Types'), 'hr-resource-types-list', $options),
       'separator' => TRUE,
     ],
     [
       'permissions' => ["administer users", "access users overview"],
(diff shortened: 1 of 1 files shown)


civihr_default_theme/tests/TemplateTest.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+class TemplateTest extends PHPUnit_Framework_TestCase {
```
davialexandre: Thanks for adding this, but I'm afraid we'll have to keep the application without test coverage for this for now. This type of test is more suited for an e2e test suite, which, unfortunately, we still don't have for CiviHR. If we leave this test here, people will think this is the right way to write these tests and will start duplicating them. CiviHR will have e2e tests eventually (it has been discussed multiple times), and when that happens there will probably be test cases to cover this. For now, please remove the tests you've added.

davialexandre (changes requested):
@mickadoo You're referring to link using two different names and none of them are the actual one. In the title you're using "Onboarding Administration", in the description, you're using "Administer Onboarding Wizard", but the images show a "Customize Welcome Wizard" link. Please use the correct name in all places.

civihr_default_theme/tests/TemplateTest.php
mickadoo: Your'e right, this test does suit e2e, at least for the happy case of ensuring the menu items appear. 

For edge cases (checking permissions etc.) I'd still say there's value in having unit tests. I don't think testing functions like this is the best way though. I would prefer not to require calling Drupal functions for a unit test, although it will require some bigger adjustments in this repo before we can use unit tests with mocking since there are no classes at all. I'll remove the tests for now anyway.

davialexandre (approved):

PR #301, opened 2018-01-22
PCHR-3202: SSP Tasks Styling Fixes

deb1990:
## Overview
In SSP - Tasks and Assignments the following is broken,
1. 'Show completed tasks' is not aligned as in SSP for staff
2. Search icon is partly hidden behind 'Enter name' search field and is not aligned

## Before
**Desktop**
tasks___civihrnew2

**Mobile**
tasks___civihrnew2

## After
**Desktop**
tasks___civihrnew2

**Mobile**
tasks___civihrnew2

## Technical Details
This is a regression issue caused by https://github.com/compucorp/civihr-employee-portal-theme/commit/45184bf2f8ec0a32469fc826809ec88ec82ab87a

Which has changes from PCHR-1251(L&A), so when we merged L&A into staging, this got broken.

Added necessary css changes to fix the styles.

---

- [x] Backstop Tests Pass


igorpavlov-zz:
@deb1990 please specify which exact tests were run and passed.

civihr_default_theme/scss/components/_panel.scss
```diff
@@ -152,11 +152,13 @@ form {
     padding: 0;
     position: absolute;
     right: 20px;
-    top: 17px;
+    top: 13px;
```
igorpavlov-zz: Please fix the order of the CSS rules. If this can be done in the whole file - that would be great. Please do so in a separate commit and do not mix with code changes.

civihr_default_theme/scss/components/_panel.scss
```diff
@@ -191,15 +193,18 @@ form {
...
-    padding: 6px;
+    right: 0;
+    padding: 0;
     font-size: 1.1rem;
-    line-height: 1;
+    line-height: 34px;
     background: none;
     color: $crm-color-neutral;
     cursor: default;
+    z-index: 4;
```
igorpavlov-zz: Why 4 and not 5 or 12?

igorpavlov-zz:
"Show completed tasks" is still one pixel too low vertically.

civihr_default_theme/scss/components/_panel.scss
```diff
@@ -152,11 +152,13 @@ form {
     padding: 0;
     position: absolute;
     right: 20px;
-    top: 17px;
+    top: 13px;
 
     > [class*="chr_panel__toolbar__"] {
       display: inline-block;
+      line-height: 34px;
```
igorpavlov-zz: I can notice 4 places in this code change where the number `34` is used. I wonder if it is a coincidence, and if it is not - the number should be stored in a variable.

igorpavlov-zz:
Could you please provide screenshots for mobile as well please?

igorpavlov-zz (changes requested):

civihr_default_theme/scss/components/_panel.scss
deb1990: Used variable `$input-height-base`.

civihr_default_theme/scss/components/_panel.scss
deb1990: tasks___civihrnew2
Because the input field has `z-index` of 3 in active state, so 4 is necessary so that the search icon gets displayed.

civihr_default_theme/scss/components/_panel.scss
```diff
@@ -199,12 +199,12 @@ form {
...
     padding: 0;
     font-size: 1.1rem;
-    line-height: 34px;
+    line-height: $input-height-base;
     background: none;
     color: $crm-color-neutral;
     cursor: default;
     z-index: 4;
-    width: 34px;
+    width: $input-height-base;
```
igorpavlov-zz: That does not really sound well, it either:
1) needs a comment OR
2) var needs to be renamed

igorpavlov-zz (changes requested):

civihr_default_theme/scss/components/_panel.scss
```diff
@@ -72,6 +72,7 @@ form {
 
   .form-type-select {
     @include fa('\f0d7');
+
     display: inline-table; // Safari *shrugs*
```
igorpavlov-zz: This comment should be enhanced and written professionally.

igorpavlov-zz (changes requested):

civihr_default_theme/scss/components/_panel.scss
deb1990: that is a default bootstrap variable, which we have used in other places also. that is the height for any form element

igorpavlov-zz (approved):

PR #307, opened 2018-01-23
PCHR-3051: Fix Webform Conditional Hiding

mickadoo:
## Overview

Webform conditionals allow you to hide / show a field based on the value of another field. However we add a [prefix](https://github.com/compucorp/civihr-employee-portal-theme/blob/9bc4b207f7a7457cf9dbfe4b9a16613dc7f18579/civihr_default_theme/template.php#L752) to the fields that contains the label. This is not hidden as part of the webform conditionals, which leaves just the label visible if a field is hidden. This PR fixes that.

## Before

Webform conditionals would not hide the label for a field if it was hidden by the conditional logic. Also we used a Javascript function to get around this for the NI/SSN field in the onboarding form.

## After

The class that the webform conditionals use is added to the wrapper div. This means that the entire element (including the label) will be hidden if the webform conditional logic defines that it should be.

## Comments

`hideSSNLabelOnCheckboxClick` is no longer necessary because of the fix. However since [two conditionals cannot target the same element](https://www.drupal.org/project/webform/issues/2481043) and the specs define that NI/SSN must be hidden _both_ when the "Skip this step" box is checked _and_ when the "I am applying for NI/SSN" box is checked we need to use a separate Javascript function to handle the hiding of this element. I just repurposed `hideSSNLabelOnCheckboxClick` for this.

---

- [ ] Tests Pass
No tests in this repo

### DIFF
--- a/civihr_default_theme/assets/js/civihr_default_theme.script.js
+++ b/civihr_default_theme/assets/js/civihr_default_theme.script.js
@@ -53,7 +53,7 @@
       addVerticalLineInCustomizeOnboardingPage();
       applyCustomSelectOnRadioClick();
       handleWebformCalendar();
-      hideSSNLabelOnCheckboxClick();
+      hideSSNFieldOnSkipCheckboxClick();
       removeTextFromCarouselPager();
       Drupal.civihr_theme.createDragAndDrop('.onboarding_wizard_profile_pic_upload_image input[type="file"]');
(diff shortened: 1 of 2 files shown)


tunbola (approved):

civihr_default_theme/assets/js/civihr_default_theme.script.js
```diff
@@ -275,14 +275,16 @@
   /**
    * Hide the SSN Label of Onboarding page when the checkbox is checked
```
deb1990: Please update the description accordingly.

deb1990 (changes requested):

PR #316, opened 2018-02-09
PCHR-3259: Use lint staged for pre-commit linters

AkA84:
See https://github.com/civicrm/civihr/pull/2464 

## Additional changes
There were two `.gitignore` files, one in the repo root and one under the civihr_default_theme/ folder. This PR merges them into one file, in the repo root

Note: there are also two `package.json` files, but i decided to keep them separated because the one in civihr_default_theme/ is part of the auto-generated theme from radix, so I didn't want to disrupt the original structure. What should really happen is to move the entire theme directly under the root repo (really not sure why it was put in a sub-dir in the first place), but I didn't want to increase the scope of this PR, which is already big enough (it already spans all 6 repos)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
 *.DS_Store
 *.css.map
-
+*.codekit
+.idea
+.sass-cache
+.DS_Store
+bower_components
+node_modules
+extensions/compass_radix
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@
 .DS_Store
 bower_components
 node_modules
-extensions/compass_radix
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,9 @@
 *.DS_Store
 *.css.map
-
+*.codekit
+.idea
+.sass-cache
+.DS_Store
+bower_components
+node_modules
+extensions/compass_radix
```
igorpavlov-zz: Why doesn't this ignore have "/*/"?

igorpavlov-zz (approved):

PR #328, opened 2018-03-19
PCHR-3426: Fix buttons and button groups styling

igorpavlov-zz:
## Overview

This PR applies a rule for large buttons. This PR is a part of https://github.com/civicrm/civihr/pull/2536. Note, there was no usage of `btn-lg` in the L&A before https://github.com/civicrm/civihr/pull/2536.

Initially, the button had fixed height as 40px defined by a style CiviHR. The font size remained the same. This style was removed and the button class has been amended with `btn-lg`. However, the font-size has been increased because this is a default Bootstrap behaviour.

SSP Theme does not have buttons with increased font-size, but has buttons with increased height, such as a button described above. This is why `btn-lg` in SSP should increase the height, but not the font-size.

## Before

(!) This is not how it currently looks on Staging, but this is how it looks once https://github.com/civicrm/civihr/pull/2536 is applied.

The first button is inside a `btn-group` container. The second one does not have a container. Both buttons have `btn-lg` class.

image

## After

The font size is not changed, only the size of the button. This is consistent with the current state 

image

## Technical description

```css
.btn-lg {
  font-size: inherit;
```

The component should have 20px bottom margin. There is no relevant variable holding this value.

```css
leave-request-record-actions {
  margin-bottom: 20px;
```

### Removing shadows from buttons

image

```
/sites/all/themes/civihr_employee_portal_theme/civihr_default_theme/assets/bower_components/bootstrap-sass/assets/stylesheets/bootstrap/_buttons.scss
```

```css
.btn {
  ...

  &:active,
  &.active {
    ...
    @include box-shadow(inset 0 3px 5px rgba(0,0,0,.125));
```

In SSP no buttons should ever have shadows.

```css
.btn {
  &:active,
  &.active {
    box-shadow: none;
```

## Comments

This PR also disables the rule to force exactly one space after `:` in SASS and updates 'stylelint-scss' to 2.5.0 version.

---- 

✅Backstop tests - passed
✅CSS distributive files - included


civihr_default_theme/scss/components/_button.scss
```diff
@@ -2,7 +2,11 @@
   border: transparent;
   padding: 0 10px;
 
-  & + .dropdown-menu > li > a {
+  + .dropdown-menu a {
```
AkA84, line 14: why this change? There is no mention in the PR

civihr_default_theme/scss/components/_button.scss
```diff
@@ -2,7 +2,11 @@
...
     color: $brand-primary;
   }
 }
+
+.btn-lg {
```
AkA84: You are overriding a basic bootstrap style, but I cannot find any explanation for why this was necessary in the PR. You are basically altering the look of a variation of a fundamental bootstrap component, and this must then be very well justified

AkA84 (commented):
Couple of comments

1) You are changing a (fundamental) style, but you didn't include any screenshot
2) You didn't justify the change
3) You mention L&A, but it's not clear why that is. What does have L&A to do with it? Why is it influencing the SSP theme at such a fundamental level?
4) The code snippet doesn't really serve any purpose here. There is nothing valuable in having bits (or the entirety) of your solution copy/pasted in the PR description if there is no particular reason to do so (for example if you need to document a particular aspect of your solution), as a reviewer I really gain nothing in terms of understanding from that snippet there, I could easily see it in the diffs view. It would have been far more valuable in this case to provide a screenshot instead of the code

civihr_default_theme/scss/components/_button.scss
igorpavlov-zz: To satisfy the linter.

civihr_default_theme/scss/components/_button.scss
```diff
@@ -2,7 +2,11 @@
...
+  font-size: inherit;
```
AkA84: please don't use `inherit`, but use the `font-size-base` variable, like in shoreditch

AkA84 (changes requested):

civihr_default_theme/scss/components/_button.scss
igorpavlov-zz: Done.

civihr_default_theme/scss/components/_button.scss
igorpavlov-zz: In Shoreditch the large button has the base font size, while in original Bootstrap the font is bigger for larger buttons. SSP should be consistent with Shoreditch in this particular point, I think, because there are elements reused in both Admin and SSP and they should look the same.

Shoreditch:
image

SSP before fix:
image

SSP after fix:
image

civihr_default_theme/scss/components/_button.scss
```diff
@@ -1,8 +1,16 @@
+.btn {
+  box-shadow: none !important;
```
AkA84: i don't see any explanation for this in the PR. Why is this necessary? I cannot see any `box-shadow` applied on buttons in the SSP

igorpavlov-zz: @AkA84 updated the description and the commit. https://github.com/compucorp/civihr-employee-portal-theme/pull/328/commits/b4ee86a531ab577e62c44f53d07dc53ef5efcef0#diff-df5d683b3db579634d916ca4a287a0dfR1

AkA84 (approved):

PR #332, opened 2018-03-29
PCHR-3426: Add SSP secondary buttons style

igorpavlov-zz:
## Overview

This PR adds styles to secondary buttons `.btn-secondary`.

## Before

image

## After

image

## Technical Details

```js
// Variables file
$brand-secondary:     #727e8a !default;
// Buttons component
.btn-secondary {
  @include button-variant($light, $brand-secondary, $btn-secondary-border);
}
```

## Comments
1. Variables file is refactored
2. Common file is refactored
3. Buttons styles moved from common file to the buttons component file

⚠️The linting of `common.scss` file during PR Feedback commit was ignored. Reasons are:
1) This file is very old, not well organised and large
2) Some rules would be very hard to fix, such as:
```css
select[multiple],
select[size] {
  height: auto;
}
```
3) This file is aimed to disappear at some point anyway due to the development of new SSP theme


civihr_default_theme/scss/base/_custom.scss
```diff
@@ -1,3 +1,5 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id, scss/at-extend-no-missing-placeholder */
```
AkA84: i don't know why you are using this, you should use it only for the civicrm/ part of shoreditch

AkA84 (changes requested):

AkA84 (approved):

PR #341, opened 2018-05-15
(Hotfix)FAR-263: Fix Broken Yoti Icon

deb1990:
## Overview
After updating Yoti module to latest 7.x-1.9, the yoti button is broken on Welcome Page. This PR fixes that.

## Before
2018-05-15 at 4 52 pm

## After
2018-05-15 at 4 51 pm

## Technical Details
1. The style got broken because in 7.x-1.9 the path to the Yoti SDK JS file has changed from
`https://sdk.yoti.com/clients/browser.js` to `https://sdk.yoti.com/clients/browser.2.0.1.js`.

So instead of just updating the URL, the `YotiHelper.php` file from the Yoti module is imported, and the `YOTI_SDK_JAVASCRIPT_LIBRARY` constant is used to always get latest URL.

---


hitesh-jain:
Looks good. @deb1990

civihr_default_theme/template.php
```diff
@@ -481,10 +485,14 @@ function civihr_default_theme_js_alter(&$javascript) {
...
       'sites/all/themes/civihr_employee_portal_theme/civihr_default_theme/assets/js/radix.modal.js',
       'sites/all/modules/civicrm/packages/jquery/plugins/jquery.blockUI.min.js',
-      'https://sdk.yoti.com/clients/browser.js',
       'sites/all/modules/civihr-signup/js/civihr-signup.js',
       'https://use.typekit.net/mhr5yod.js',
     ];
+
+    if (module_exists('yoti')) {
+      array_push($requiredJS, YotiHelper::YOTI_SDK_JAVASCRIPT_LIBRARY);
+    }
```
davialexandre: There's no need for calling `module_exists` twice. Please move the require_once from line 17 to this if. Also, the usual way to add an item to the end of an array in PHP is to use the `[]` operator, not `array_push`:

```php
$requiredJS[] = YotiHelper::YOTI_SDK_JAVASCRIPT_LIBRARY;
```

davialexandre (changes requested):

davialexandre (approved):

PR #363, opened 2018-06-14
PCHR-3835: Hide Text format

deb1990:
## Overview
Text format section is hidden in Edit HR Document screen.

## Before
2018-06-14 at 6 41 pm

## After
2018-06-14 at 6 41 pm


beto-aveiga (approved):
From a developer perspective It would be more appropriate to remove this field from the server side, from a hook_form_alter, but looking at the SCSS code I see that there is another line that does the same thing with styling, which makes me think it is a common practice, that is why I'm approving.

PR #369, opened 2018-07-04
PCHR-3887: Sync styles

igorpavlov-zz:
## Overview

This PR syncs styles with https://github.com/compucorp/civihr/pull/2739.


civihr_default_theme/scss/base/_custom.scss
```diff
@@ -628,12 +628,6 @@ h2.pane-title {
 // Dropdowns
 
 .dropdown-menu {
-  padding: 0;
-
-  > li > a {
```
AkA84: as discussed in the call, in order to make the smallest change that would still make things look decent enough, please remove only the `padding` for `.dropdown-menu`, keep the style of `> li > a`

AkA84 (changes requested):

civihr_default_theme/scss/base/_custom.scss
AkA84: also, did you give a quick look around the SSP to find other places where we have `.dropdown-menu` and how this change affects them?

igorpavlov-zz: I searched through all CiviHR by 'dropdown-menu' in '*.html' files and found out that only these dropdowns will be affected in SSP:

image

image

The rest is outside SSP.

AkA84 (approved):

PR #366, opened 2018-06-26
PCHR-3834: Update Global Styles

deb1990:
## Overview
As part of the PR the Global Styles has been updated of the SSP theme according to the new mockup.
Related PRs
https://github.com/compucorp/civihr-employee-portal/pull/502
https://github.com/compucorp/org.civicrm.shoreditch/pull/112

#### Brand Colors
```scss
$brand-primary: #0071bd !default;
$brand-secondary: #9494a5 !default;
$brand-success: #44cb7e !default;
$brand-warning: #e6ab5e !default;
$brand-danger: #cf3458 !default;
```
#### Gray Colors
```scss
$gray-darker: #3e3e54 !default;
$gray-dark: #464354 !default;
$gray: #4d4d69 !default;
$gray-light: #9494a5 !default;
$gray-lighter: #c2cfd8 !default;

$light: #fff !default;
$light-pale: #e8eef0 !default;
$light-gray: #f3f6f7 !default;
```
Other than the bootstrap variables, one more gray color is found in the mockup, it is mentioned as
```scss
$chr-gray-1: #4e4e78;
```
In future if other colors are required, the same format needs to be followed

#### Gradients
For the Save button the following gradient variables are created
```scss
$chr-gradient-1: linear-gradient(315deg, #01A2D5 0%, #0070BB 100%);
$chr-gradient-2: linear-gradient(315deg, #0187B1 0%, #005B97 100%);
```
The above gradients will be used in multiple places, for example 
Borders in https://projects.invisionapp.com/d/main#/console/13248050/297993949/preview
In Payroll Details https://projects.invisionapp.com/d/main#/console/13248050/297230123/preview

So there is no known pattern at this moment. So for simplicity's sake, the names are numbered. Future gradients should be added similarly.

#### Fixing Adhoc Colors
Previously throughout the theme, many adhoc colors were used, all of those are removed and variables from `scss/base/_variables.scss` has been used.

#### Font Size
The new font sizes are mapped to bootstrap variables
```scss
$font-size-base: 13px;
$font-size-small: 13px;
$font-size-large: 18px;

$font-size-h1: 24px;
$font-size-h2: 24px;
$font-size-h3: 24px;
$font-size-h4: $font-size-large;
$font-size-h5: $font-size-large;
$font-size-h6: $font-size-large;
```
And a new variable is created for font-size, which could not be mapped to bootstrap
```scss
$chr-font-size-medium: 14px;
```

#### Font Weights
Variables has been created/updated for Font weights
```scss
[...]

PR #371, opened 2018-07-05
PCHR-3934: Move and style HR Admin Button for small Devices

deb1990:
## Overview
As part of this PR the "HR Admin" button is moved to the bottom of the main menu in SSP and changes its style for the mobile viewport.

## Before
2018-07-05 at 3 56 pm

## After
2018-07-05 at 3 55 pm

## Technical Details 
 - The Desktop menu uses `flexbox` styling so that the position of the items can be swapped from css using `order` property. Same is implemented for smaller devices.

## Testing
Tested manually on 
- Chrome (macOS)
- Mobile Emulator in Chrome (macOS)


igorpavlov-zz:
@deb1990 May we amend the Overview description a bit?

```
This PR moves the HR Admin button to the bottom of the main menu in SSP and changes its style for the mobile viewport.
```

igorpavlov-zz:
@deb1990 please add the information on how this was tested. I think you cannot really test it with Backstop (yet), but did you manually test it? Different browsers, OSs, devices etc. Please provide this information.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -2,7 +2,7 @@
   @include clearfix;
 
   background: $gray-dark;
-  box-shadow: 0 2px 3px 0 rgba(black, 0.15);
+  box-shadow: 0 2px 3px 0 rgba(#000, 0.15);
```
igorpavlov-zz: I think we have a variable for black:

`$crm-black: #000 !default;`

civihr_default_theme/scss/components/_header.scss
```diff
@@ -106,12 +104,12 @@
 
 .chr_header__logo {
   color: $light;
-  font-size: 1.6rem;
   display: inline-block;
+  font-size: 1.6rem;
```
igorpavlov-zz, line 108: Can we have a variable for this value?

civihr_default_theme/scss/components/_header.scss
```diff
@@ -106,12 +104,12 @@
...
 }
 
 .chr_header__nav {
-  background: #ffffff;
+  background: #fff;
```
igorpavlov-zz: `$crm-white: #fff !default;`

civihr_default_theme/scss/components/_header.scss
```diff
@@ -173,14 +172,16 @@
...
-        border: none;
+        border: 0;
       }
     }
   }
 
-  @media(min-width: $screen-md-max) {
-    display: flex;
+  .menu-link-hr-admin {
+    order: 9999;
```
igorpavlov-zz: Please add a comment why 9999, something like `to ensure it appears the last in the list`

civihr_default_theme/scss/components/_header.scss
```diff
@@ -173,14 +172,16 @@
...
+  }
 
+  @media(min-width: $screen-md-max) {
```
igorpavlov-zz: We mix `min-width` rule with `max` variable. Could you check that on this exact edge case (when the screen has **exactly** `$screen-md-max` width), it behaves as expected?

civihr_default_theme/scss/components/_header.scss
```diff
@@ -218,6 +218,19 @@
   }
 }
 
+@media (max-width: $screen-md-max - 1) {
```
igorpavlov-zz: This also does not sound right to me. If let's say `$screen-md-max === 500`, then it means that `500` would still be **screen-md**. In this rule you say `for screen that is less or equal 499`.

Am I missing something?

Also, not sure why do you need this rule, you can simply position the *default* rules and then override with media query below. You don't need two media queries.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -218,6 +218,19 @@
...
+  .chr_header__nav__menu {
+    flex-direction: column;
+  }
+
+  .menu-link-hr-admin {
+    border: 1px solid $gray-darker;
+    line-height: 50px;
```
igorpavlov-zz: Please put this value into a variable.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -218,6 +218,19 @@
...
+    margin: 25px;
```
igorpavlov-zz: Please put this value into a variable.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -264,8 +277,8 @@
 
   a {
     color: $light-pale;
-    font-size: 1.3rem;
     display: block;
+    font-size: 1.3rem;
```
igorpavlov-zz, line 282: Please put this value into a variable.

igorpavlov-zz (changes requested):

igorpavlov-zz:
@deb1990 Let's attempt to put any values that are not really self descriptive into variables. Ideally, every value should be justified by design instructions.

Please note, these are absolutely fine:
```css
width: 0 // no width
float: right // attach to right
width: 100% // full width
display: none // do not show
// etc
```

cc @AkA84 @reneolivo

civihr_default_theme/scss/components/_header.scss
deb1990: Thanks for the suggestion. In SSP it's called `$light` though, so using the same.

civihr_default_theme/scss/components/_header.scss
deb1990: Good point, fixed all such instances in the code.

civihr_default_theme/scss/components/_header.scss
deb1990: After correcting all variables in the file as you mentioned, `-1` is not necessary anymore.

> Also, not sure why do you need this rule, you can simply position the default rules and then override with media query below. You don't need two media queries.

Done

deb1990:
@igorpavlov I have only added variables for the places you have mentioned. But skipped the font related comments. The whole SSP theme is supposed to undergo redesign, which has already started as part of https://github.com/compucorp/civihr-employee-portal-theme/pull/366. We are already moving all font sizes to variables in that PR.

>Let's attempt to put any values that are not really self descriptive into variables. Ideally, every value should be justified by design instructions.

Yes we need to have a bigger discussion about this, because throughout SSP we have many such values. So I suggest to do this as part of global styling updates i.e. PCHR-3676.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -1,8 +1,12 @@
+$chr-hr-admin-mobile-line-height: 50px;
```
igorpavlov-zz: We do `mobile first` so it would be great if variables are named in an opposite way:

```css
$chr-hr-admin-line-height: 50px;
$chr-hr-admin-margin:      25px;
$chr-hr-admin-desktop-margin:             33px;
```

Also, I think there is a slight problem with self-explanation of vars names, I read `$chr-hr-admin-margin` as a margin of an admin. I think it lacks an explanation of `margin of what?`. Is it a margin of a button?

civihr_default_theme/scss/components/_header.scss
```diff
@@ -124,17 +128,19 @@
...
   a {
-    border-bottom: 1px solid $crm-grizzly;
-    color: $gray-darker;
-    display: block;
-    margin: 0 25px;
-    padding: 0 20px;
+    border: 0;
+    color: inherit;
+    margin: 0;
+    padding: 0 10px;
```
igorpavlov-zz, line 130: Why the padding has changed from 20px to 10px? Is it an instruction from a designer or does it just look better?

Also, please have a look if we already have a variable for `0 10px` and what is it for. If there is no such a var, then please consider creating it.

civihr_default_theme/scss/components/_header.scss
```diff
@@ -178,35 +184,47 @@
   }
 
   .menu-link-hr-admin {
-    order: 9999;
+    line-height: $chr-hr-admin-margin;
+    order: 9999; // to ensure it appears the last in the list
+    padding: 12px 0 0 15px;
```
igorpavlov-zz: Same, needs a variable I think. Also, where does this value come from? If **not** from a designer, can we contact them and ask what the padding should be?

civihr_default_theme/scss/components/_header.scss
```diff
@@ -178,35 +184,47 @@
...
+
     a {
-      border: 0;
-      color: inherit;
-      margin: 0;
-      padding: 0 10px;
+      border-bottom: 1px solid $crm-grizzly;
+      color: $gray-darker;
+      display: block;
+      margin: 0 25px;
```
igorpavlov-zz, line 201: Same, should be a variable.

igorpavlov-zz (changes requested):

deb1990:
@igorpavlov as per our discussion I have removed the variables.

civihr_default_theme/scss/components/_header.scss
deb1990: > Why the padding has changed from 20px to 10px? Is it an instruction from a designer or does it just look better?

Previously the styles for mobile was declared here, and for desktop we had `@media(min-width: $screen-md-max) {` block below. But as per your previous comment(which I agree with) we should define the desktop rules first and have media queries for other devices. So I have moved around the styles, but the end result is still the same.

deb1990: Also, the value is not used anywhere in this partial. So not creating variables.

reneolivo (approved):

PR #372, opened 2018-07-09
PCHR-3937: Style changes for the Main Menu in smaller devices

deb1990:
## Overview
As part of this PR the following changes had been implemented
1. The styles for the selected item of the list has been changed.
2. Removed the styles to highlight HOME in grey.
3. Hiding the HR ADMIN button in mobile devices(Smaller than 768px in width).
4. In Tablets the HR ADMIN button is made smaller and center aligned.

## Before
**Mobile**
2018-07-09 at 3 52 pm

**Tablets**
2018-07-09 at 3 53 pm

## After
**Mobile**
2018-07-09 at 3 54 pm

**Tablets**
2018-07-09 at 6 32 pm

## Tests Done
Tested manually on

Chrome (macOS)
Mobile and Tablet Emulator in Chrome (macOS)


civihr_default_theme/scss/components/_header.scss
```diff
@@ -215,16 +221,23 @@
     .menu-link-hr-admin {
       border: 1px solid $gray-darker;
       line-height: 50px;
-      margin: 25px;
+      margin: 25px auto;
       padding: 0;
       text-align: center;
+      width: 300px;
```
igorpavlov-zz: I don't think we should set a fixed width, can we make it `display: inline-block` instead?

civihr_default_theme/scss/components/_header.scss
```diff
@@ -194,8 +189,19 @@
...
     flex-direction: column;
 
-    a {
+    > li {
       border-bottom: 1px solid $crm-grizzly;
+
+      &.active { /* stylelint-disable-line selector-no-qualifying-type */
+        background-color: $gray-light;
+
+        &::after { /* stylelint-disable-line max-nesting-depth */
```
igorpavlov-zz: Would `&.active::after` require `stylelint-disable-line max-nesting-depth`?

igorpavlov-zz (changes requested):

civihr_default_theme/scss/components/_header.scss
deb1990: Removing the width completely works too. Please check the updated screenshots.

civihr_default_theme/scss/components/_header.scss
deb1990: Yes. Fixed.

igorpavlov-zz (approved):

PR #373, opened 2018-07-13
PCHR-3360: Fix Users List date filters

deb1990:
## Overview
Last Access From and To Filters are not working if just one of the fields is selected on Manage Users. This PR and https://github.com/compucorp/civihr-employee-portal/pull/514 fixes that.

## Before
before1

## After
after1

## Technical Details
1. In https://github.com/compucorp/civihr-employee-portal/pull/514 it is documented, that we are now using bootstrap grids to form the exposed filter section. So all the css previously written to create custom grid has been removed.

2. Previously there a single exposed filter with "Is between" operator. But "Is between" does not work when one of the conditions are empty, so instead of 1 filter, added two filters
    - Is greater than or equal to
    - Is less than or equal to

      for the same reason, id's has been updated in `civihr_default_theme.script.js`.

## Comments
Linting errors fixed in the touched SCSS files.
 - `@extend civihr_form__fieldset;` has been removed, instead `civihr_form__fieldset` is added to the markup whenever `civihr_form__fieldset--transparent` is used.
 -  Styles for `#civihr-employee-portal-document-form` and `.civihr_form--modal` is removed from SCSS files, as it was not used.

---

- [x] Manual Tests Pass


civihr_default_theme/assets/js/civihr_default_theme.script.js
```diff
@@ -326,9 +328,9 @@
     }
 
     var calendarIcon = '<span class="input-group-addon pointer"><i class="fa fa-calendar"></i></span>';
-    $('#edit-access-min').datepicker();
-    $('#edit-access-min').after(calendarIcon);
-    $('#edit-access-max').datepicker();
-    $('#edit-access-max').after(calendarIcon);
+    $('#edit-access-from').datepicker();
```
AkA84, line 331: why are the ids different? Were they wrong to begin with? Were they changed in this PR? The description says nothing about it

civihr_default_theme/scss/drupal/_form.scss
```diff
@@ -44,23 +44,39 @@
...
+    width: auto;
+  }
+
+  .views-widget-filter-name,
+  .views-widget-filter-rid,
+  .views-widget-filter-access {
+    padding-right: 20px;
+  }
+
+  .views-widget-filter-access,
```
AkA84: what are all the new styles? The description doesn't mention them

civihr_default_theme/scss/drupal/_form.scss
```diff
@@ -145,18 +161,20 @@
...
-      &:before {
+      &::before {
         content: '\f073';
         font-family: FontAwesome;
         position: absolute;
         right: -20px;
         top: 7px;
       }
 
+      /* stylelint-disable selector-max-compound-selectors */
```
AkA84, line 141: can't we refactor the style instead

civihr_default_theme/scss/pages/_users-list.scss
```diff
@@ -19,19 +19,19 @@
...
       border: 0;
-      color: #fff;
+      color: $light;
       display: inline-block;
       margin-left: 20px;
       padding: 0.7rem 1.1rem;
     }
   }
+
+  #edit-actions-submit {
```
AkA84, line 32: why is this new style needed? No mention in the PR description

AkA84 (changes requested):

civihr_default_theme/scss/pages/_users-list.scss
deb1990: This is not new, its moved to avoid more than 3 nested selectors.

civihr_default_theme/scss/drupal/_form.scss
deb1990: Explained in the description.

civihr_default_theme/scss/drupal/_form.scss
deb1990: I have refactored the styles which could be, but where it could not be done because of the required structure, linting has been disabled. This is similar.

civihr_default_theme/assets/js/civihr_default_theme.script.js
deb1990: Explained in the description.

AkA84 (approved):

PR #380, opened 2018-08-15
PCHR-3681: Remove hardcoded paths

davialexandre:
## Overview

A site created with the CiviHR installation profile will have the modules, extensions, and themes located in different folders than sites created with the other installation scripts. This theme uses some resources with some hardcoded paths following the folder structure used in existing sites and these fail to load in a site created with the profile.

## Before

The resources used by the theme fail to load in a site built using the profile.

## After

The resources used by the theme load sucessfully in sites built with the profile and with buildkit/civihr-installer.

## Technical Details

Every Drupal site has a `profiles` folder, where all the profiles are kept. Everything used by the profile (modules, libraries and themes) are kept inside the profile own folder instead of the site's global folder. To better understand the difference, let's consider the `civihr_employee_portal_theme` theme as an example:

In a site created with the existing installation scripts, this theme will be located at `sites/all/themes/civihr_employee_portal_theme`, while in a site created with the profile, it will be located at `profiles/civihr/themes/civihr_employee_portal_theme`.

To fix this issue and make sure the resources will always be loaded no matter where they are located, the hardcoded paths have been removed and the `drupal_get_path()` is used to return the paths for modules and themes.

### DIFF
--- a/civihr_default_theme/template.php
+++ b/civihr_default_theme/template.php
@@ -368,12 +368,16 @@ function civihr_default_theme_css_alter(&$css) {
     $requiredCSS = [
       'modules/system/system.base.css',
       'modules/system/system.theme.css',
-      'sites/all/modules/civicrm/bower_components/font-awesome/css/font-awesome.css',
-      'sites/all/modules/civihr-contrib-required/fancy_login/css/fancy_login.css',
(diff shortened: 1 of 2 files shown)


ajesamson (approved):
Approved

PR #388, opened 2018-10-31
PCHR-4355: Fix gulp sass

AkA84:
This PR simply updates the theme's css files based on the changes done in https://github.com/compucorp/org.civicrm.shoreditch/pull/135


civihr_default_theme/assets/css/civihr_default_theme.style.css
```diff
@@ -1,4 +1,4 @@
....input-group-btn>select.chr_action--icon--edit[multiple],[data-ssp-upload-document] .input-group-sm>.input-group-btn>select.chr_action--icon--list[multiple],[data-ssp-upload-document] .input-group-sm>.input-group-btn>select.chr_action--icon--print[multiple]{height:auto}.form-group-sm .form-control,.form-group-sm .ui-datepicker .ui-datepicker-header select,.ui-datepicker .ui-datepicker-header .form-group-sm select{height:30px;padding:5px 10px;font-size:12px;line-height:1.5;border-radius:0}.form-group-sm select.form-control,.form-group-sm .ui-datepicker .ui-datepicker-header select,.ui-datepicker .ui-datepicker-header .form-group-sm select{height:30px;line-height:30px}.form-group-sm textarea.form-control,.form-group-sm select[multiple].form-control,.form-group-sm .ui-datepicker .ui-datepicker-header select[multiple],.ui-datepicker .ui-datepicker-header .form-group-sm select[multiple]{height:auto}.form-group-sm .form-control-static{height:30px;min-height:32px;padding:6px 10px;font-size:12px;line-height:1.5}.input-lg,.input-group-lg>.form-control,.ui-datepicker .ui-datepicker-header .input-group-lg>select,.input-group-lg>.input-group-addon,.input-group-lg>.input-group-btn>.btn,.input-g
```
igorpavlov: I have a feeling that something may be wrong here, it has 2 rules for `border-color`, could you please doublecheck that the dist file changes are expected?:

image

igorpavlov-zz, line 1: I have a feeling that something may be wrong here, it has 2 rules for `border-color`, could you please doublecheck that the dist file changes are expected?:

image

igorpavlov-zz (changes requested):

civihr_default_theme/assets/css/civihr_default_theme.style.css
igorpavlov-zz: I know this is not due to your change, but because I can now spot it, I decided to raise it.

AkA84: Was due to the fact that the `button-outline-variant` mixin is used, and that the `border-color` and `color` properties were then overridden
```scss
.btn-default-outline {
  @include button-outline-variant($crm-white);
  border-color: $crm-white !important;
  color: $crm-white !important;
}
```
(from https://github.com/compucorp/org.civicrm.shoreditch/pull/21)

Problem was that those overrides were pointless, both `border-color` and `color` were already assigned `$crm-white` by the mixin, and I don't see any benefit of redefine them with an `!important`. So i've removed then, thanks for spotting it

igorpavlov-zz (approved):

civihr_default_theme/assets/css/civihr_default_theme.style.css
igorpavlov-zz: I was asking about duplicating the same style when there is a non-important rule and the same, important rule. I guess that's just how the compiler works, it simply does not care.

However, the change you have just introduced requires Backstop tests I think.

igorpavlov-zz (changes requested):

civihr_default_theme/assets/css/civihr_default_theme.style.css
igorpavlov-zz: The reply to this comment is here: https://github.com/compucorp/org.civicrm.shoreditch/pull/135#discussion_r229676147

igorpavlov-zz (approved):

PR #390, opened 2018-11-27
PCHR-4427: Upgrade npm dependencies

AkA84:
Upgrading npm dependencies to patch security vulnerability issues.

The `gulp-imagemin` package was bumped to a new major version
```diff
- "gulp-imagemin": "^4.1.0",
+ "gulp-imagemin": "^5.0.3",
```

thus `gulp images` (the task using the package) was ran to check if there were any differences in the output.

Some of the images were indeed reported as different by git (hence their inclusion in this PR), although they did not display any discernible visual difference when checked with a tool like https://diffchecker.com/image-diff

Example

**Before**
welcome-img-before

**After**
welcome-img-after


igorpavlov-zz:
Could you include 2 screenshots of versions of a sample image (before/after)?

AkA84:
@igorpavlov done

civihr_default_theme/assets/images/spinner.svg
```diff
@@ -1,13 +1 @@
...
-    d="M40,72C22.4,72,8,57.6,8,40C8,22.4,
-       22.4,8,40,8c17.6,0,32,14.4,32,32c0,1.1-0.9,2-2,2
-       s-2-0.9-2-2c0-15.4-12.6-28-28-28S12,24.6,12,40s12.6,
-       28,28,28c1.1,0,2,0.9,2,2S41.1,72,40,72z">
-  </path>
-  <title>
-    CiviHRSpinner
-  </title>
-</svg>
+<svg xmlns="http://www.w3.org/2000/svg" data-name="Layer 1" viewBox="0 0 80 80"><path fill="#0071BD" d="M40 72C22.4 72 8 57.6 8 40S22.4 8 40 8s32 14.4 32 32c0 1.1-.9 2-2 2s-2-.9-2-2c0-15.4-12.6-28-28-28S12 24.6 12 40s12.6 28 28 28c1.1 0 2 .9 2 2s-.9 2-2 2z"/></svg>
```
igorpavlov-zz, line 1: Probably need an end line

igorpavlov-zz (changes requested):

civihr_default_theme/assets/images/spinner.svg
igorpavlov-zz: image

igorpavlov-zz (approved):

igorpavlov-zz:
@AkA84 it seems that every time we run `gulp images` it will remove the end line from the svg file (maybe from other files as well). This should be checked so it does not break things.

AkA84:
@igorpavlov that's correct, so there's no point trying to add the new line manually. I've forced pushed the svg file as created by the gulp task

